PR #109, opened 2019-10-27
Feature/crs matrix setters

rrsettgast:


src/src/CRSMatrixView.hpp
```diff
@@ -261,6 +261,83 @@ class CRSMatrixView : protected SparsityPatternView<COL_TYPE, INDEX_TYPE>
     return nRemoved;
   }
 
+  inline
+  void setEntries( T const & rhs )
```
corbett5: Assuming that `T` is POD this is fine, but if it's something like an `std::string` then this would access some uninitialized values. It would be safest to iterate over the rows in a double loop.

rrsettgast: The problem I was thinking about was that when we call an insert type method, it can resize (not reallocate), and then access something that isn't set by this function?

corbett5: But when you call insert you have to specify a new value.

rrsettgast:
@corbett5 can you take this commit and add it to your refactor PR?

rrsettgast:
These changes were committed in a separate PR

PR #114, opened 2019-11-18
guarded FPE exceptions. removed some signals from handler

rrsettgast:
This PR fixes the incorrect handling of certain system signals.

This should fix some inexplicable runtime errors we see on some platforms.


rrsettgast:
> I've had problems with `feenableexcept( ... | FE_INVALID );` on Ascent (and therefore potentially on Summit). I don't know if there's a specific macro guard though. If we ever have to deal with it again, we can add a cmake flag in the future.

Yeah. This was causing issues on Lassen/Sierra. We will have to set the FPE on Power9/NVIDIA systems in a different way.

PR #117, opened 2019-12-03
Feature/corbett/cxx cleanup

corbett5:


src/ChaiBuffer.hpp
```diff
@@ -49,14 +48,18 @@ static std::mutex chaiLock;
...
+    size = std::to_string( bytes >> 20 ) + "MB";
   }
   else
   {
-    return std::to_string(bytes >> 10) + "KB";
+    size = std::to_string( bytes >> 10 ) + "KB";
   }
+
+  // return a space padded size
+  return std::string( 7 - size.length(), ' ' ) + size;
```
rrsettgast: huh? why 7?

src/IntegerConversion.hpp
```diff
@@ -0,0 +1,150 @@
...
+ * @param input  the integer to convert
+ * @return the converted integer
+ *
+ * This function takes in an unsigned integer and converts it to a signed integer. There is a check
+ * to make sure that no data is lost as a result of this conversion.
+ */
+template< typename RTYPE, typename T >
+inline typename std::enable_if< std::is_unsigned< T >::value && std::is_signed< RTYPE >::value,
+                                RTYPE >::type
+integer_conversion( T input )
```
rrsettgast: is this actually a new file?

src/Macros.hpp
```diff
@@ -0,0 +1,181 @@
+/*
```
rrsettgast: and this?

rrsettgast: also....is this no longer dependent on slick?

src/ChaiBuffer.hpp
corbett5: That gives you five digit characters in the `XXXXXMB` string where X is either a digit or leading space. That lets you get up to transfers of 99999MB or 97GB which I think is a reasonable upper limit for now.

src/IntegerConversion.hpp
corbett5: No. I think if you look commit by commit you'll see that it got moved but that when GitHub merges all the changes together it thinks it's new.

src/Macros.hpp
corbett5: Correct.

src/IntegerConversion.hpp
corbett5: No. If you look commit by commit it should show that it was moved but when GitHub displays all the changes at once it things its a new file.

PR #119, opened 2019-12-16
Add is_base_of

AntoineMazuyer:
To go with PR https://github.com/GEOSX/GEOSX/pull/672


src/templateHelpers.hpp
```diff
@@ -80,4 +80,10 @@ struct is_instance_of : std::is_same< T, U > {};
 template< typename T, typename U >
 constexpr bool is_instance_of_v = is_instance_of< T, U >::value;
 
+template <class T, class U>
+struct is_base_of: std::is_base_of< T, U > {};
+
+template <class T, class U>
+constexpr bool is_base_of_v = is_base_of<T, U>::value;
```
corbett5: `is_base_of_v` is fine because it's C++17 but I'd get rid of `is_base_of` since it's a simple wrapper.

rrsettgast: Is there something wrong with just using `std::is_base_of`? I never really had a big problem with the `::value` specifier. Either way, I would git rid of the wrapper.

AntoineMazuyer: @klevzoff maybe you have an input on that ? To be honest I don't have the answer to this question.

AntoineMazuyer: @klevzoff maybe you have an answer to that. To be honest I don't have the answer.

klevzoff: I only added it following an example of `is_instance_of` (which is just `std::is_same` but with the ability to do specializations should we ever need to). There's nothing wrong with just using `std::is_base_of<>`.

corbett5: Oh yeah we should remove `is_instance_of` and `is_instance_of_v` and just use `std::is_same`  and also defining `is_same_v`.

corbett5: That's pretty trivial since it's only used in a few places but also kinda outside this PR. If you want to do it go for it.

AntoineMazuyer: Is it ok for me to merge the PR now ? and then maybe we can do this in a separate PR?

corbett5: Could you remove `is_base_of`? And then in the future I'll remove `is_instance_of`.

corbett5: Awesome, thanks a bunch. Merge whenever.

PR #121, opened 2019-12-20
Feature/corbett/benchmarks

corbett5:


corbett5:
See benchmarks/results.txt to get started.

corbett5:
> 
> 
> Yeah. I can't follow this at all.

I went through a bunch of iterations on how to best do this and I wound up using a bunch of macros. I think at this point it's pretty easy for me to extend the framework and add more benchmarks, but I'd like to document it so that other people can add them as well.

The crux of the problem is that I want to test every combination of the following `[native for-loop, raja serial, raja omp, raja cuda] x [ array fortran accessor, array subscript accessor, raja view, raw pointer ] x [ permutations ]`. This lends itself naturally to templates but I think the benchmarks are more reliable if the kernels are in a separate translation unit so that's why I have the `benchmarkXXXKernels.hpp` and the `benchmarkXXXKernels.cpp` where I define and explicitly instantiate the kernels. Furthermore gbenchmark doesn't support typed tests like gtest does so I need to define a benchmark for each template combination this is what led to the creation of the `FIVE_BENCHMARK_TEMPLATES_TWO_TYPES` macro.

PR #122, opened 2019-12-20
Feature/wrtobin/gpu packing

wrtobin:


src/Array.hpp
```diff
@@ -540,7 +540,10 @@ class Array : public ArrayView< T,
   std::enable_if_t< isArray< U > >
   move( chai::ExecutionSpace const space, bool const touch=true )
   {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
```
rrsettgast: Why is this pragma required? What is the effect with the presence of the horrid `reinterpret_cast`?

corbett5:
Can you squash all these commits into one?

PR #124, opened 2020-01-07
Simplified Array to string logic.

corbett5:


src/ArrayUtilities.hpp
```diff
@@ -34,7 +34,7 @@ namespace cxx_utilities
 
 /**
  * @struct stringToArrayHelper
- * This is a helper struct to recursivly read an istringstream into an array.
+ * This is a helper struct to recursively read an istringstream into an array.
```
rrsettgast: spelling schmelling.

src/ArrayUtilities.hpp
```diff
@@ -300,92 +300,47 @@ static void stringToArray( LvArray::Array< T, NDIM, PERMUTATION, INDEX_TYPE, DAT
...
+ * @param view the view to output.
+ * @return a reference to the ostream.
  */
 template< typename T,
           int NDIM,
           int UNIT_STRIDE_DIM,
           typename INDEX_TYPE,
           template< typename > class DATA_VECTOR_TYPE >
-static std::string arrayToString( LvArray::ArrayView< T, NDIM, UNIT_STRIDE_DIM, INDEX_TYPE, DATA_VECTOR_TYPE > const & array )
+std::ostream & operator<<( std::ostream & stream,
```
rrsettgast: so are the stream operators replacing the `arrayToString, stringToArray` functions in the code?...and do we need error checking for this operation? I know I did it somewhere...maybe in the xml reader. but it should be here.

corbett5: Not `stringToArray`, just `arrayToString` which shouldn't need any error checking.

PR #127, opened 2020-01-25
Fixes bug regarding NewChaiBuffer pointer record allocators.

corbett5:


src/NewChaiBuffer.hpp
```diff
@@ -157,6 +157,11 @@ class NewChaiBuffer : public chai::CHAICopyable
     newRecord->m_size = newCapacity * sizeof( T );
     newRecord->m_user_callback = m_pointer_record->m_user_callback;
 
+    for( int space = chai::CPU ; space < chai::NUM_EXECUTION_SPACES ; ++space )
```
corbett5: @rrsettgast This is the actual fix.

src/MallocBuffer.hpp
```diff
@@ -59,20 +59,20 @@ class MallocBuffer : public bufferManipulation::VoidBuffer
    */
   LVARRAY_HOST_DEVICE RAJA_INLINE constexpr
   MallocBuffer( bool=true ):
-    m_capacity( 0 ),
-    m_data( nullptr )
+    m_data( nullptr ),
```
rrsettgast: is seems like you are just changing the order here? Does this change behavior?

corbett5: No, but it's probably good policy to have the pointer as the first member. Then effectively to index into the array the compiler does `reinterpret_cast<T*>(*this)[ i ]` instead of `reinterpret_cast<T*>(*(this + member_offset))[ i ]`. It almost certainly doesn't make a difference but it won't hurt.

corbett5:
> 
> 
> Just a general question I should have asked before, but why is NewChaiBuffer/MallocBuffer needed? ArrayView no longer derives from ManagedArray. What are we gaining here?

`ArrayView` never derived from `chai::ManagedArray`. 

We're not using the `MallocBuffer` anywhere but it allows us to have a functioning buffer if we ever want to turn off `chai` and it provides a good reference for further buffer implementations.

`NewChaiBuffer` (I'll rename it at some point) doesn't use the `chai::ManagedArray` but it provides the same functionality by using the `chai::ArrayManager` directly. The benefits of this include that it is half the size, we have direct control over the behavior and I believe we can use chai develop instead of a branch with a few changes. It performed substantially better in some of the benchmarks.

rrsettgast:
> > Just a general question I should have asked before, but why is NewChaiBuffer/MallocBuffer needed? ArrayView no longer derives from ManagedArray. What are we gaining here?
> 
> `ArrayView` never derived from `chai::ManagedArray`.
> 
> We're not using the `MallocBuffer` anywhere but it allows us to have a functioning buffer if we ever want to turn off `chai` and it provides a good reference for further buffer implementations.
> 
> `NewChaiBuffer` (I'll rename it at some point) doesn't use the `chai::ManagedArray` but it provides the same functionality by using the `chai::ArrayManager` directly. The benefits of this include that it is half the size, we have direct control over the behavior and I believe we can use chai develop instead of a branch with a few changes. It performed substantially better in some of the benchmarks.

yeah. duh. it never derived. It used to have a `chai::ManagedArray`.

performed better in the accessor benchmarks? Why is it half the size? I agree it is better to direct control over behavior, but is it functionally different than the `chai::ManagedArray`?

corbett5:
> 
> yeah. duh. it never derived. It used to have a `chai::ManagedArray`.
> 
> performed better in the accessor benchmarks? Why is it half the size? I agree it is better to direct control over behavior, but is it functionally different than the `chai::ManagedArray`?

The functionality is identical to the `ChaiBuffer` which does use the `chai::ManagedArray`.

These are the members of `chai::ManagedArray`
```c++
mutable T* m_active_pointer = nullptr;
mutable T* m_active_base_pointer = nullptr;
ArrayManager* m_resource_manager = nullptr;
size_t m_elems = 0;
size_t m_offset = 0;
PointerRecord* m_pointer_record = nullptr;
bool m_is_slice = false;
```

These are the members of `NewChaiBuffer`
```c++
T * restrict m_pointer = nullptr;
std::ptrdiff_t m_capacity = 0;
chai::PointerRecord * m_pointer_record = nullptr;
```

And yes it performed better in the accessor benchmarks.

rrsettgast:
> > yeah. duh. it never derived. It used to have a `chai::ManagedArray`.
> > performed better in the accessor benchmarks? Why is it half the size? I agree it is better to direct control over behavior, but is it functionally different than the `chai::ManagedArray`?
> 
> The functionality is identical to the `ChaiBuffer` which does use the `chai::ManagedArray`.
> 
> These are the members of `chai::ManagedArray`
> 
> ```c++
> mutable T* m_active_pointer = nullptr;
> mutable T* m_active_base_pointer = nullptr;
> ArrayManager* m_resource_manager = nullptr;
> size_t m_elems = 0;
> size_t m_offset = 0;
> PointerRecord* m_pointer_record = nullptr;
> bool m_is_slice = false;
> ```
> 
> These are the members of `NewChaiBuffer`
> 
> ```c++
> T * restrict m_pointer = nullptr;
> std::ptrdiff_t m_capacity = 0;
> chai::PointerRecord * m_pointer_record = nullptr;
> ```
> 
> And yes it performed better in the accessor benchmarks.

Well then do you want to deprecate `ChaiBuffer`?

corbett5:
> Well then do you want to deprecate `ChaiBuffer`?

At the moment it's not being used anywhere outside of tests. I'd like to keep it in the repo for a bit but eventually assuming there's not reason to go back I'll remove it.

rrsettgast:
> > Well then do you want to deprecate `ChaiBuffer`?
> 
> At the moment it's not being used anywhere outside of tests. I'd like to keep it in the repo for a bit but eventually assuming there's not reason to go back I'll remove it.

Maybe stick it in a `deprecated` folder so it is clear what is being actively used.

PR #129, opened 2020-02-06
Feature/corbett/flipping arrays

corbett5:


benchmarks/benchmarkArray1DR2TensorMultiplicationKernels.cpp
```diff
@@ -66,11 +66,11 @@ namespace benchmarking
   RAJA_OUTER_LOOP( INNER_LOOP( a_ijl, b_ilk, c_ijk ) )
 
 
-template< int UNIT_STRIDE_DIM >
+template< int USD >
```
rrsettgast: US Dollars? 😄

src/ArraySlice.hpp
```diff
@@ -280,6 +280,82 @@ class ArraySlice
...
+ * @brief Apply the funtion @p f to the value @p value also
+ *        passing @p f any indices used to reach @p value.
+ * @param slice the slice to iterate over.
+ * @param f the lambda to apply to each value.
+ * @param indices the previous sliced off indices.
+ */
+DISABLE_HD_WARNING
+template< typename T, typename LAMBDA, typename ... INDICES >
+LVARRAY_HOST_DEVICE
+void forValuesInSliceWithIndices( T & value, LAMBDA && f, INDICES const ... indices )
```
rrsettgast: I am a little confused by the indices parameter. In the calls to this function, it appears that this parameter is not given an argument?? The `indices` arguments that I see in the call are arguments to the lambda, not this function?

corbett5: Correct, this is the base case for the following function. This allows you to do something like
```c++
Array< int, 3 > x( 2, 3, 4 );
Array< int, 2 > y( 2, 3, 4 );
...
// Adds x_ijk to y_ijk
forValuesInSliceWithIndices( x, 
  [&]( int const & value, auto const... indices )
  {
    y( indices... ) += value;
  }
);
```
Which is pretty trivial, but when the loop is in a templated method taking that can take in an arbitrary dimension array with any permutation it is very useful.

src/ArrayView.hpp
```diff
@@ -412,36 +412,19 @@ class ArrayView
...
-    for( INDEX_TYPE i = 0 ; i < stride0 ; ++i )
-    {
-      dest[i] = source[i];
-    }
+    forValuesInSliceWithIndices( (*this)[ sourceIndex ],
+                                 [destIndex, this]( T const & sourceVal, auto const ... indices )
+      {
+        (*this)( destIndex, indices ... ) = sourceVal;
+      }
+                                 );
```
corbett5: @rrsettgast

PR #132, opened 2020-02-28
Bugfix/missing metis neighbors

corbett5:


src/ArrayOfSets.hpp
```diff
@@ -422,13 +422,15 @@ class ArrayOfSets : protected ArrayOfSetsView< T, INDEX_TYPE >
...
      * @brief Callback signaling that the size of the set has increased.
+     * @param [in] curPtr the current pointer to the array.
      * @param [in] nToAdd the increase in the size.
      * @note This method doesn't actually change the size, it just checks if the new size
      *       exceeds the capacity of the set and if so reserves more space.
      * @return a pointer to the sets values.
      */
     inline
-    T * incrementSize( INDEX_TYPE const nToAdd ) const LVARRAY_RESTRICT_THIS
+    T * incrementSize( T * const LVARRAY_UNUSED_ARG( curPtr ),
```
rrsettgast: can you change `LVARRAY_UNUSED_ARG` to `LVARRAY_UNUSED_PARAM`?...
and why add this parameter if it us unused?

src/ArrayOfSetsView.hpp
```diff
@@ -300,7 +300,7 @@ class ArrayOfSetsView : protected ArrayOfArraysView< T, INDEX_TYPE, std::is_cons
     ARRAYOFARRAYS_CHECK_BOUNDS( i );
 
     INDEX_TYPE const setSize = sizeOfSet( i );
-    T const * const setValues = (*this)[i];
+    T * const setValues = getSetValues( i );
```
rrsettgast: How did this work before when `setValues` was a pointer to const?

src/ArrayOfSets.hpp
corbett5: It's used in the default `sortedArrayManipulation::CallBack::incrementSize`.

src/ArrayOfSetsView.hpp
corbett5: The `setValues` is used to determine the number of insertions that take place, then the `incrementSize` call back returns the pointer that is used to do the insertions.

PR #133, opened 2020-02-28
Add data() method to ArraySlice

klevzoff:
Related to https://github.com/GEOSX/GEOSX/pull/740

### DIFF
--- a/src/ArraySlice.hpp
+++ b/src/ArraySlice.hpp
@@ -246,6 +246,13 @@ public:
     return m_data == ptr;
   }
 
+  /**
+   * @brief Return a pointer to the values.
+   */
+  LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
+  T * data() const
+  { return m_data; }
+
 #if defined(USE_TOTALVIEW_OUTPUT) && !defined(__CUDA_ARCH__) && defined(USE_ARRAY_BOUNDS_CHECK)
   /**
    * @brief Static function that will be used by Totalview to display the array contents.
--- a/src/ArraySlice.hpp
+++ b/src/ArraySlice.hpp
@@ -167,8 +167,8 @@ public:
   operator[]( INDEX_TYPE const index ) const noexcept LVARRAY_RESTRICT_THIS
   {
     ARRAY_SLICE_CHECK_BOUNDS( index );
-    return ArraySlice< T, NDIM-1, USD-1, INDEX_TYPE >( m_data + ConditionalMultiply< 0, USD >::multiply( index,
-                                                                                                         m_strides[ 0 ] ), m_dims + 1,
+    return ArraySlice< T, NDIM-1, USD-1, INDEX_TYPE >( m_data + ConditionalMultiply< 0, USD >::multiply( index, m_strides[ 0 ] ),
+                                                       m_dims + 1,
                                                        m_strides + 1 );
   }
 
@@ -247,11 +247,53 @@ public:
   }
 
   /**
+   * @brief Check if the slice is contiguous in memory
+   * @tparam _USD dummy template parameter equal to USD; do not replace
+   * @return @p true if represented slice is contiguous in memory
+   */
+  template< int _USD = USD >
+  LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
+  std::enable_if_t< ( _USD >= 0), bool >
+  isContiguous() const
+  {
+    bool rval = true;
+    for( int i = 0 ; i < NDIM ; ++i )
+    {
+      if( i == _USD ) continue;
+      INDEX_TYPE prod = 1;
+      for( int j = 0 ; j < NDIM ; ++j )
+      {
+        if( j != i ) prod *= m_dims[j];
+      }
+      rval &= (m_strides[i] <= prod);
+    }
+    return rval;
+  }
+
+  /**
+   * @brief Check if the slice is contiguous in memory
+   * @tparam USD_ dummy template parameter equal to USD; do not replace
+   * @return @p false, this overload is enabled for slices that
+   *         have already lost its unit stride dimension
+   */
+  template< int USD_ = USD >
+  LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
+  std::enable_if_t< (USD_ < 0), bool >
+  isContiguous() const
+  {
+    return false;
+  }
+
+  /**
    * @brief Return a pointer to the values.
+   * @pre The slice must be contiguous
    */
   LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
   T * data() const
-  { return m_data; }
+  {
+    LVARRAY_ASSERT( isContiguous() );
+    return m_data;
+  }
 
 #if defined(USE_TOTALVIEW_OUTPUT) && !defined(__CUDA_ARCH__) && defined(USE_ARRAY_BOUNDS_CHECK)
   /**
--- a/src/ArraySlice.hpp
+++ b/src/ArraySlice.hpp
@@ -289,9 +289,9 @@ public:
    * @pre The slice must be contiguous
    */
   LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
-  T * data() const
+  T * dataIfContiguous() const
   {
-    LVARRAY_ASSERT( isContiguous() );
+    LVARRAY_ERROR_IF( !isContiguous(), "The slice must be contiguous for direct data access" );
     return m_data;
   }
 
--- a/unitTests/testArraySlice.cpp
+++ b/unitTests/testArraySlice.cpp
@@ -35,9 +35,42 @@ void resizeArray( Array< T, NDIM, PERM, INDEX_TYPE > & a, DIMS const & newdims,
 template< typename PERMUTATION >
 class PermutedSliceTest : public ::testing::Test {};
 
-TYPED_TEST_CASE_P( PermutedSliceTest );
+using Permutations = ::testing::Types< RAJA::PERM_IJ,
+                                       RAJA::PERM_JI,
+                                       RAJA::PERM_IJK,
+                                       RAJA::PERM_JIK,
+                                       RAJA::PERM_IKJ,
+                                       RAJA::PERM_KIJ,
+                                       RAJA::PERM_KJI,
+                                       RAJA::PERM_JKI,
+                                       RAJA::PERM_IJKL,
+                                       RAJA::PERM_JIKL,
+                                       RAJA::PERM_IKJL,
+                                       RAJA::PERM_KIJL,
+                                       RAJA::PERM_JKIL,
+                                       RAJA::PERM_KJIL,
+                                       RAJA::PERM_IJLK,
+                                       RAJA::PERM_JILK,
+                                       RAJA::PERM_ILJK,
+                                       RAJA::PERM_LIJK,
+                                       RAJA::PERM_JLIK,
+                                       RAJA::PERM_LJIK,
+                                       RAJA::PERM_IKLJ,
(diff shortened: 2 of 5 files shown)


src/ArraySlice.hpp
```diff
@@ -246,6 +246,13 @@ class ArraySlice
     return m_data == ptr;
   }
 
+  /**
+   * @brief Return a pointer to the values.
+   */
+  LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
+  T * data() const
+  { return m_data; }
```
klevzoff: This is needed in order to pass 2D slices to linear algebra TPLs during assembly, since 2D slices are not implicitly convertible to a pointer

corbett5: This can lead to out of bounds access when the slice's data isn't contiguous (comes from a permuted higher dimensional array). Compile time checking for this isn't straightforward since `ArraySlice` only knows the unit stride dimension and not the permutation of the parent `Array`.

For instance a 2D slice of an `Array< int, 3, RAJA::PERM_JIK>` isn't contiguous but has the same unit stride dimension as a 2D slice from the standard layout `Array< int, 3, RAJA::PERM_IJK >`.

You could add in a runtime check to verify that the `ArraySlice` holds contiguous data, but I'd prefer you just pass `&array(0, 0)` at the call site.

rrsettgast: @klevzoff Are there cases where you are are not using stack arrays to hold the local data?

klevzoff: @rrsettgast Not in function that I wrote, I don't think. But need to go through all solvers to find out.

rrsettgast: I see what you mean. 

I think there should be some functionality for direct data access (i.e. `data()`). However using the pointer returned from `data()` is error prone. Using `&array[0][0]` is just as error prone but makes the developer work harder to get the data. So options:

1. Provide `data()` method that returns a pointer to the data. Let the user shoot themselves in the foot.
2. Make the user get the pointer by `&array[0][0]`. Let the user shoot themselves in the foot.
3. Provide some sort of controlled access to the data pointer. For instance, in the case of an `ArraySlice2d`, there could be a SFINAE enabled function that provides access row major access and column major access:
```
template< _NDIM=NDIM, _USD=USD >
LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
typename std::enable_if< _NDIM==2 && _USD==1, T *>::type 
data1() const { return m_data; }

template< _NDIM=NDIM, _USD=USD >
LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
typename std::enable_if< _NDIM==2 && _USD==0, T *>::type 
data0() const { return m_data; }
```
With option 3, the "user" requests a pointer to 2d array data with the explicit expectation of the data permutation. Of course this would be extendible to `NDIM>2` since we could do a check to make sure that the strides are consistently decreasing/increasing from right to left indices.

corbett5: I still prefer option 2 since it is visible at the call site that the user is doing something unusual. My point was that you can have `NDIM == 2` and `USD == 1` and the data still won't be contiguous. Even if it is contiguous it might not be in the order you expect it to be in. The only safe solution I see is to do a runtime calculation of the strides from the dimensions and see if they match the actual strides.

rrsettgast: @corbett5

I would argue for some variant option 3 based on the frequency of the known usage. This approach will be used everywhere in the `LAI` where we interface with underlying `TPL` functions. It would be nice to localize the checks somewhere...either in the ArraySlice or in a helper function.

```
template< _NDIM=NDIM, _USD=USD >
LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
typename std::enable_if< _NDIM==2 && _USD==1, T *>::type 
data1() const
{
  LVARRAY_ASSERT( m_strides[0]==m_dims[1] ); 
  return m_data;
}

template< _NDIM=NDIM, _USD=USD >
LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
typename std::enable_if< _NDIM==2 && _USD==0, T *>::type 
data0() const
{
  LVARRAY_ASSERT( m_strides[1]==m_dims[0] ); 
  return m_data;
}
```

corbett5: @klevzoff do you pass the 2D slice to a method or do you just need to pass the data?
``` c++
Array3d< int > x;

f( x[ 0 ] );
// or 
f( x[ 0 ].data(), x[ 0 ].size() );
```

If it's the second option we could add a method in `ArrayView` that was conditioned on the standard layout that returns a pointer to the given slice.

``` c++
Array3d< int > x;
f( x.data( 0 ), x.size() / x.size( 0 ) );
```

klevzoff: @corbett I'm passing the slice to the LAI method, which under the hood then passes the raw pointer/size to the library.

To be clear, I don't (currently) actually have use cases where I would need to pass a slice of a higher-dimensional array, although nothing prevents the user from doing so. Slice is only used because it's the only unifying type between chai-based arrays, stack-based arrays, malloc-based arrays, etc. etc. The most common, by far, use case is:
```c++
stackArray1d<globalIndex,N> indicies(n);
stackArray2d<real64,N*N> values(n,n);
...
matrix.add(indices, indices, values); // implicitly converts to arraySlice2d<real64,1>
```

rrsettgast: I think the problem is that, as @klevzoff mentioned, is that these methods don't know where the `arraySlice2d<T,USD>` is coming from. Thus, as @corbett5 mentioned, there is no guarantee that the caller of `EpetraMatrix::add()` provides an `ArraySlice` that is contiguous. So the proposed call to get the pointer (`values.data()`) should check to make sure that `values` is in the layout that `m_epetra_matrix->add()` expects.

corbett5: So I'd be all for adding a method `bool ArraySlice::isContiguous()` and then in then
``` c++
void EpetraMatrix::add( arraySlice1d<real64> rowIndices,
                        arraySlice1d<real64> colIndices,
                        arraySlice2d<real64, 0> values )
{
  GEOSX_ASSERT( values.isContiguous() );
  m_epetra_matrix->add( rowIndices.size()
                        rowIndices, // implicit conversion
                        colIndices.size()
                        colIndices, // implicit conversion
                        values.data(), // obtain raw ptr somehow
                        Epetra::COLUMN_MAJOR ); // tell it about ordering
}
```

rrsettgast:
```
template< _NDIM=NDIM, _USD=USD >
LVARRAY_HOST_DEVICE inline
typename std::enable_if< _USD==0, T *>::type 
bool ArraySlice::isContiguous() const
{
  bool rval = true;
  localIndex thisStride = 1;
  for( int dim=0 ; dim<_NDIM ; ++dim )
  {
    if( thisStride != m_stride[dim] )
    {
      rval = false;
      break;
    }
    thisStride *= m_dims[dim];
  }
}

template< _NDIM=NDIM, _USD=USD >
LVARRAY_HOST_DEVICE inline
typename std::enable_if< _USD==_NDIM-1, T *>::type 
bool ArraySlice::isContiguous() const
{
  for( int dim=_NDIM-1 ; dim>=0; ++dim )
  {
    if( thisStride != m_stride[dim] )
    {
      rval = false;
      break;
    }
    thisStride *= m_dims[dim];
  }
}

LVARRAY_HOST_DEVICE  inline T *
dataContiguous() const
{
  LVARRAY_ASSERT( isContiguous() ); 
  return m_data;
}
```

or something like this...perhaps the implementation is incorrect.

corbett5: I bet there's an implementation that doesn't need to specialization.

klevzoff: Well... the slice is NOT contiguous whenever at least one of the strides is greater than the product of other dims.. I think? I don't know if that's the most optimal check to do, but I can write it.

rrsettgast: @klevzoff By no means will I claim that this the best check...Just an example. In general, what I put up mimics the way the strides are calculated using the dims....we can probably just reuse the stride calculator for this....and makes sure that the contiguous strides are equal to `m_strides`.

klevzoff: I took a stab at it... but it's Friday night and my brain is likely slightly dysfunctional.

src/ArraySlice.hpp
```diff
@@ -246,6 +246,55 @@ class ArraySlice
...
+    for( int i = 0 ; i < NDIM ; ++i )
+    {
+      if( i == _USD ) continue;
+      INDEX_TYPE prod = 1;
+      for( int j = 0 ; j < NDIM ; ++j )
+      {
+        if( j != i ) prod *= m_dims[j];
+      }
+      rval &= (m_strides[i] <= prod);
+    }
```
klevzoff: All of this should ideally be a compile-time recursion... but I tried and it looked ugly

corbett5: Yeah I like this better.

src/ArraySlice.hpp
```diff
@@ -246,6 +246,55 @@ class ArraySlice
...
+  }
+
+  /**
+   * @brief Return a pointer to the values.
+   * @pre The slice must be contiguous
+   */
+  LVARRAY_HOST_DEVICE inline CONSTEXPRFUNC
+  T * data() const
+  {
+    LVARRAY_ASSERT( isContiguous() );
```
corbett5: Could you rename this to `dataIfContiguous` and the `LVARRAY_ASSERT` to `LVARRAY_ERROR_IF`?

unitTests/testArraySlice.cpp
```diff
@@ -0,0 +1,112 @@
...
+                                         RAJA::PERM_JKLI,
+                                         RAJA::PERM_KJLI,
+                                         RAJA::PERM_JLKI,
+                                         RAJA::PERM_LJKI,
+                                         RAJA::PERM_KLJI,
+                                         RAJA::PERM_LKJI >;
+
+INSTANTIATE_TYPED_TEST_CASE_P( 2D, PermutedSliceTest, Permutations2D );
+INSTANTIATE_TYPED_TEST_CASE_P( 3D, PermutedSliceTest, Permutations3D );
+INSTANTIATE_TYPED_TEST_CASE_P( 4D, PermutedSliceTest, Permutations4D );
```
corbett5: Could you change these to just standard typed tests? There not a reason to use the parameter version here and that's what the other tests use (or should use).

klevzoff: There is a somewhat weak reason. Because the following is not legal:
```
TYPED_TEST_CASE( PermutedSliceTest, Permutations2D );
TYPED_TEST_CASE( PermutedSliceTest, Permutations3D );
TYPED_TEST_CASE( PermutedSliceTest, Permutations4D );
```
It works if I put all permutations (2D/3D/4D) into a single type list, e.g.
```
using Permutations = ::testing::Types< RAJA::PERM_IJ, ..., ..., RAJA::PERM_LKJI >;
TYPED_TEST_CASE( PermutedSliceTest, Permutations );
```
which slightly loses on expressiveness, because the test names you see at runtime are something like `PermutedSliceTest/camp::idx_seq<0, 1, 2>/IsContiguous>`, so it's a bit more difficult to decipher the dimension, compared to having it right in the test name. Although it's a pretty minor detail.

In general I tend to prefer _type-parametrized_ tests to _typed_ tests, because they don't require pre-defining a global list of types to be tested. In cases like LAI this is the only option, since the list of types changes based on which packages you compile with, so a global type tuple is (nearly) impossible to define. In this case it doesn't really matter that much, so I'll do it.

corbett5: That's sounds reasonable to me, but I see you already made the change. Oh well.

PR #136, opened 2020-03-12
Changes for SFINAE consistency.

corbett5:


src/ArrayOfArraysView.hpp
```diff
@@ -148,7 +148,7 @@ class ArrayOfArraysView
    *        isn't invoked, this usually occurs with template argument deduction.
    */
   LVARRAY_HOST_DEVICE CONSTEXPRFUNC inline
-  ArrayOfArraysView< T, INDEX_TYPE const, true > const & toViewC() const LVARRAY_RESTRICT_THIS
+  ArrayOfArraysView< T, INDEX_TYPE const, true > const & toViewSemiConst() const LVARRAY_RESTRICT_THIS
```
rrsettgast: `toViewSemiConst` is a little strange sounding. Would it make sense to give this a more descriptive name? For instance `toViewConstSizes()`?

src/CRSMatrixView.hpp
```diff
@@ -116,7 +116,7 @@ class CRSMatrixView : protected SparsityPatternView< COL_TYPE, INDEX_TYPE >
    *        when the above UDC isn't invoked, this usually occurs with template argument deduction.
    */
   LVARRAY_HOST_DEVICE CONSTEXPRFUNC inline
-  CRSMatrixView< T, COL_TYPE const, INDEX_TYPE const > const & toViewC() const LVARRAY_RESTRICT_THIS
+  CRSMatrixView< T, COL_TYPE const, INDEX_TYPE const > const & toViewSemiConst() const LVARRAY_RESTRICT_THIS
```
rrsettgast: Same question as above. Proposing something like `toViewConstSparsity()`?

src/ArrayOfArraysView.hpp
corbett5: Yeah it's not a great name. It might be useful in the future to have a name that's consistent across the different classes, but `toViewConstSparsity` is a good for the `CRSMatrix` class.

src/NewChaiBuffer.hpp
```diff
@@ -286,23 +333,23 @@ class NewChaiBuffer : public chai::CHAICopyable
...
 
-  template< typename U=T >
-  std::enable_if_t< !std::is_base_of< CHAICopyable, U >::value >
-  moveInnerData( chai::ExecutionSpace const LVARRAY_UNUSED_ARG( space ) )
+  template< typename U=T_non_const >
+  std::enable_if_t< !internal::HasMemberFunction_move< U > >
+  moveInnerData( chai::ExecutionSpace const, std::ptrdiff_t const, bool const )
   {}
 
-  template< typename U=T >
```
corbett5: @rrsettgast This is a bug that I believe is in develop. There are two issues here first is that in the for-loop below you loop up to the capacity of the buffer, which means you might be accessing uninitialized objects if the size is less than the capacity. The second is that if you have `array1d< SortedArray< T > >` and you call move on it then you'll be calling the `SortedArray` copy constructor which does a deep copy.

PR #141, opened 2020-04-15
Feature/corbett/blueprint

corbett5:


src/ArrayView.hpp
```diff
@@ -23,15 +23,16 @@
 #ifndef ARRAYVIEW_HPP_
 #define ARRAYVIEW_HPP_
 
-// Source includes
+/// Source includes
```
rrsettgast, line 26: Does doxygen actually pick this up?

src/NewChaiBuffer.hpp
```diff
@@ -253,9 +253,10 @@ class NewChaiBuffer
 
     if( prevSpace == chai::CPU ) moveInnerData( space, size, touch );
 
-    m_pointer = static_cast< T * >( internal::getArrayManager().move( const_cast< T_non_const * >( m_pointer ),
-                                                                      m_pointer_record,
-                                                                      space ) );
+    const_cast< T * & >( m_pointer ) =
```
rrsettgast, line 256: I wonder if this sort of thing should be wrapped up un a function somewhere. Perhaps something like 
```
template< typename T > 
std::remove_const<T> &  breakConstCorrectness( T & constObject )
{
  return const_cast< std::remove_const<T> & >( constObject );
}
```
but something that would actually work. I think this has some issues with reference to pointer case.

PR #142, opened 2020-04-24
Changed sorted insert/remove methods to take iterators.

corbett5:


corbett5:
I need to update the documentation.

src/Array.hpp
```diff
@@ -269,7 +269,7 @@ class Array : public ArrayView< T,
     static_assert( check_dim_type< INDEX_TYPE, DIMS... >::value, "arguments to Array::resize(DIMS...newDims) are incompatible with INDEX_TYPE" );
 
     INDEX_TYPE const oldSize = size();
-    dimUnpack( const_cast< INDEX_TYPE * >( m_dims ), newDims ... );
+    dimUnpack( m_dims, newDims ... );
```
rrsettgast, line 272: when did `m_dims` become an array of non-const? Since the accessors are const, it is OK to not have `ArrayView::m_dims` non-const? Did you run any performance testing for this?

src/SortedArray.hpp
```diff
@@ -227,40 +208,23 @@ class SortedArray : protected SortedArrayView< T, INDEX_TYPE >
    * @return True iff the value was actually removed.
    */
   inline
-  bool erase( T const & value ) LVARRAY_RESTRICT_THIS
+  bool remove( T const & value ) LVARRAY_RESTRICT_THIS
```
rrsettgast, line 211: Why the name change? `erase` is the same as `std::set`...not that this is a good reason to keep `erase`.

src/SparsityPatternView.hpp
```diff
@@ -240,46 +240,18 @@ class SparsityPatternView : protected ArrayOfSetsView< COL_TYPE, INDEX_TYPE >
    *       If USE_ARRAY_BOUNDS_CHECK is defined a lack of space will result in an error,
    *       otherwise the values in the subsequent row will be overwritten.
    */
+  template< typename ITER >
   LVARRAY_HOST_DEVICE inline
-  INDEX_TYPE_NC insertNonZeros( INDEX_TYPE const row, COL_TYPE const * cols, INDEX_TYPE const ncols ) const LVARRAY_RESTRICT_THIS
+  INDEX_TYPE_NC insertNonZeros( INDEX_TYPE const row, ITER const first, ITER const last ) const LVARRAY_RESTRICT_THIS
```
rrsettgast, line 245: Is there an advantage to the iterator interface?

unitTests/testUtils.hpp
```diff
@@ -102,14 +102,14 @@ inline void forall( INDEX_TYPE const max, LAMBDA && body )
 #endif
 
 // Comparator that compares a std::pair by it's first object.
-template< class A, class B, class COMP=std::less< B > >
+template< class A, class B, class COMP=std::less< A > >
```
rrsettgast, line 105: This isn't really a comparison of the pair, but only the first component of the pair?

src/Array.hpp
corbett5: When I was doing the array benchmarking, so yeah it didn't effect anything. Pretty much every method in `ArrayView` is const, plus whenever they're captured they're const so it shouldn't make any difference.

src/SortedArray.hpp
corbett5: `insert` and `remove` seems like a better match to me. `erase` sounds good for when you want to delete a position whereas you `remove` a value. But it's all pretty trivial.

unitTests/testUtils.hpp
corbett5: Yeah `std::pair` has builtin comparators but this is used for the `CRSMatrix` tests where you have pairs of columns and values that need to be sorted by the columns.

src/SparsityPatternView.hpp
corbett5: It lets you insert directly from a `std::set` for example, it's a lot more flexible than the pointer approach. It won't be faster than the pointer version but the hit should be negligible compared to the cost of inserting.

corbett5:
@rrsettgast originally I made these changes so that in `GEOSX` we could build up a `std::set` and then insert that into a `SortedArray` or `ArrayOfSets` all at once. Part of this also simplified the insertion logic by enforcing that all insertions of multiple elements be both sorted an unique. I didn't wind up using the iterator inserts in `GEOSX` because it was faster to build up a `std::vector` and then sort it and remove duplicates.

PR #145, opened 2020-04-30
Update BLT

klevzoff:


unitTests/testBuffers.cpp
```diff
@@ -657,7 +657,7 @@ using BufferTestWithReallocTypes = ::testing::Types< NewChaiBuffer< int >,
                                                      MallocBuffer< TestString >
                                                      >;
 
-TYPED_TEST_CASE( BufferTestWithRealloc, BufferTestWithReallocTypes );
+TYPED_TEST_SUITE( BufferTestWithRealloc, BufferTestWithReallocTypes, );
```
corbett5, line 660: Interesting, does the new `TYPED_TEST_SUITE` do anything useful?

klevzoff: I don't know, but they deprecated everything `*_TEST_CASE` in a gtest release packed with BLT now, which led to clang complaining.

klevzoff: These `_TEST_SUITE` macros also have `VA_ARGS` parameters, the meaning of which doesn't seem to be documented, hence the extra commas at the end.

PR #149, opened 2020-05-07
Made the StackArray work on device.

corbett5:


src/IntegerConversion.hpp
```diff
@@ -106,17 +114,18 @@ template< typename OUTPUT, typename INPUT >
...
                   !std::is_unsigned< INPUT >::value,
                   OUTPUT >
+inline constexpr LVARRAY_HOST_DEVICE
 integerConversion( INPUT input )
 {
   static_assert( std::is_integral< INPUT >::value, "INPUT must be an integral type." );
   static_assert( std::is_integral< OUTPUT >::value, "OUTPUT must be an integral type." );
 
-  LVARRAY_ERROR_IF_LT( input, std::make_signed_t< OUTPUT >{ std::numeric_limits< OUTPUT >::min() } );
+  LVARRAY_ERROR_IF_LT( input, std::make_signed_t< OUTPUT >{ internal::IntegerTraits< OUTPUT >::min } );
```
francoishamon, line 123: The CUDA compiler on Ascent does not seem to like this `LVARRAY_ERROR_IF_LT` and fails with the error `/ccsopen/home/hamonf/GEOSX/src/coreComponents/LvArray/unitTests/../src/IntegerConversion.hpp(123): error: statement may not appear in a constexpr function`. I get the same error at line 128 of this file, and at line 84 of `StackBuffer.hpp`. When these lines are commented out I can successfully compile so this problem is not blocking.

unitTests/testStackArray.cpp
```diff
@@ -16,61 +16,265 @@
...
+  void init()
   {
-    array1d[i] = i;
-    ASSERT_TRUE( array1d.data()[i] == i );
+    int dims[ NDIM ];
+
+    for( int i = 0; i < NDIM; ++i )
+    { dims[ i ] = 8; }
+
+    m_array.resize( NDIM, dims );
```
francoishamon, line 97: In the flow solvers, there are a lot of places where we use the `stackArray` as: 
```c++
stackArray1d< real64, MAX_NUM_FACES > property( numFaces ); 
```
and with this PR tested on Ascent this is not compiling in a function marked `GEOSX_HOST_DEVICE` (and sorry if I misunderstood something). But, everything works well with a `resize`:
```c++ 
stackArray1d< real64, MAX_NUM_FACES > property;
property.resize( numFaces );
``` 
I am fine with both, but perhaps we should completely disallow the first option (on host and on device), or maybe allow it on device as well.

corbett5: Thanks, I just forgot to to decorate that method.

src/IntegerConversion.hpp
corbett5: Yeah I didn't mean to make them `constexpr`, nice catch.

unitTests/testStackArray.cpp
corbett5: @francoishamon it should work now.

PR #152, opened 2020-05-14
Add capability to restore original signal handlers

rrsettgast:


src/SetSignalHandling.cpp
```diff
@@ -27,28 +27,34 @@
 namespace LvArray
 {
 
+typedef void ( *handle_type )( int );
+static handle_type initialHandler[128] = { nullptr };
```
corbett5, line 31: I would make this a `std::map< int, handle_type >` from signal to handler so that you don't have to worry about the ordering.

rrsettgast: So much better that way.

PR #156, opened 2020-05-27
Feature/corbett/tensor ops

corbett5:


corbett5:
All the operations in `tensorOps.hpp` are tested on host and device with all combinations of c-array and `ArraySlice` arguments for a fixed set of parent `Array` permutations. Naturally there is a lot of template instantiation going on and it takes about 2 minutes to compile on Lassen. I doubt this is an issue when building GEOSX because nothing depends on it and so a thread can work on it in the background, but it could be an issue on Travis.

corbett5:
@rrsettgast Feel free to suggest new method names `LvArray::tensorOps::symmetricMatrixMatrixTranspose` is a mouthful, although if we go with the `AijBkj` method of naming I'm not sure how to distinguish it from `matrixMatrixTranspose`.

src/tensorOps.hpp
```diff
@@ -22,127 +22,934 @@
...
-  a( 1, 1 ) = temp11 * invDet;
-  a( 1, 2 ) = temp12 * invDet;
-  a( 2, 0 ) = temp20 * invDet;
-  a( 2, 1 ) = temp21 * invDet;
-  a( 2, 2 ) = temp22 * invDet;
-
-  return det;
+template< std::ptrdiff_t N, std::ptrdiff_t M, typename DST_MATRIX, typename SRC_MATRIX >
+LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK inline
+void copy( DST_MATRIX && dstMatrix, SRC_MATRIX const & srcMatrix )
```
corbett5, line 462: I intend to make all these references `restrict`.

rrsettgast: I don't know that restrict on references is widely supported? I think that gcc has some support for it, but I didn't know this was a usable feature.

rrsettgast:
> @rrsettgast Feel free to suggest new method names `LvArray::tensorOps::symmetricMatrixMatrixTranspose` is a mouthful, although if we go with the `AijBkj` method of naming I'm not sure how to distinguish it from `matrixMatrixTranspose`.

I always liked the index notation. in the case of `symmetricMatrixMatrixTranspose` and `matrixMatrixTranspose` you don't have to distinguish the function names since the parameters will be different. One will take full matrix A[3][3] and one will take symmetric matrix A[6].

src/tensorOps.hpp
corbett5: I believe both gcc and clang support it. I'll create a new macro `LVARRAY_RESTRICT_REF` so we can turn it off easily.

corbett5:
> 
> 
> > @rrsettgast Feel free to suggest new method names `LvArray::tensorOps::symmetricMatrixMatrixTranspose` is a mouthful, although if we go with the `AijBkj` method of naming I'm not sure how to distinguish it from `matrixMatrixTranspose`.
> 
> I always liked the index notation. in the case of `symmetricMatrixMatrixTranspose` and `matrixMatrixTranspose` you don't have to distinguish the function names since the parameters will be different. One will take full matrix A[3][3] and one will take symmetric matrix A[6].
> 
> However, whatever I like may not be the only consideration for this. Perhaps @castelletto1 has a suggestion?

Because the methods are all templated on the argument types you can't distinguish based on the overload.

rrsettgast:
> > > @rrsettgast Feel free to suggest new method names `LvArray::tensorOps::symmetricMatrixMatrixTranspose` is a mouthful, although if we go with the `AijBkj` method of naming I'm not sure how to distinguish it from `matrixMatrixTranspose`.
> > 
> > 
> > I always liked the index notation. in the case of `symmetricMatrixMatrixTranspose` and `matrixMatrixTranspose` you don't have to distinguish the function names since the parameters will be different. One will take full matrix A[3][3] and one will take symmetric matrix A[6].
> > However, whatever I like may not be the only consideration for this. Perhaps @castelletto1 has a suggestion?
> 
> Because the methods are all templated on the argument types you can't distinguish based on the overload.

Well....we can restrict the input types using SFINAE, but it would be better to just have different function names.

corbett5:
> Well....we can restrict the input types using SFINAE, but it would be better to just have different function names.

Yeah I'd really prefer to not do that. We could always just use different nomenclature for symmetric and non-symmetric matrices, like `AijBkj` and `SijBkj` but then to know which one to use you'd have to read the docs. Of course if you chose the wrong one you'll get a compilation error with a nice static assert mixed in with tens of other errors.

rrsettgast:
> > Well....we can restrict the input types using SFINAE, but it would be better to just have different function names.
> 
> Yeah I'd really prefer to not do that. We could always just use different nomenclature for symmetric and non-symmetric matrices, like `AijBkj` and `SijBkj` but then to know which one to use you'd have to read the docs. Of course if you chose the wrong one you'll get a compilation error with a nice static assert mixed in with tens of other errors.

Agreed. Maybe use the BLAS nomenclature?

corbett5:
> 
> 
> > > Well....we can restrict the input types using SFINAE, but it would be better to just have different function names.
> > 
> > 
> > Yeah I'd really prefer to not do that. We could always just use different nomenclature for symmetric and non-symmetric matrices, like `AijBkj` and `SijBkj` but then to know which one to use you'd have to read the docs. Of course if you chose the wrong one you'll get a compilation error with a nice static assert mixed in with tens of other errors.
> 
> Agreed. Maybe use the BLAS nomenclature?

Lets see what @castelletto1 has to say, I wouldn't mid using the BLAS names but the mapping isn't perfect.

castelletto1:
> > > > Well....we can restrict the input types using SFINAE, but it would be better to just have different function names.
> > > 
> > > 
> > > Yeah I'd really prefer to not do that. We could always just use different nomenclature for symmetric and non-symmetric matrices, like `AijBkj` and `SijBkj` but then to know which one to use you'd have to read the docs. Of course if you chose the wrong one you'll get a compilation error with a nice static assert mixed in with tens of other errors.
> > 
> > 
> > Agreed. Maybe use the BLAS nomenclature?
> 
> Lets see what @castelletto1 has to say, I wouldn't mid using the BLAS names but the mapping isn't perfect.

If we want to use the BLAS nomenclature we should have a perfect mapping. However, there's nothing wrong with using the index notation. Here are a few options:

| op | BlasLapackLA | tensorOps (a) | tensorOps (b) | tensorOps (c) | 
| --- | --- | --- |--- | --- |
| `A*B^T`    | `matrixMatrixTMultiply` | `AijBkj`| `matrixMatrixTMultiply` | `matMatTMult` | 
| `sym(A)*B^T` | none (*)  | `symAijBkj` |`symmetricMatrixMatrixTMultiply` |`symMatMatTMult` | 

If we like the option (c), we may want to conform the `BlasLapackLA` nomenclature to it.

(*) at the moment all matrix-matrix product implementation use `dgemm()`. In case the first operator is symmetric we may want to use `dsymm()`.

corbett5:
@castelletto1 I'm happy with any of those.

corbett5:
Here are some naming conventions we could use.
vectors: **x, y, z**
matrices: **A, B, C**
| Operation | Einstein notation | BlasLapackLA | current |
| - | - | - | - |
| max absolute entry | ? | `vectorNormInf`  | `maxAbsolueEntry` |
| **x**_i = **y**_i **z**_i | ? | ? | `elementWiseMultiplication` |
| **x** = **y** + **z** | ? | `vectorVectorAdd` | `add` |
| **A** = **B** + **C** | ? | `matrixMatrixAdd` | `add` |
| result = **xy**^T | `AiBi` | `vectorDot` | `innerProduct` |
| **A** = **x**^T**y** | `AiBj` | ? | `outerProduct` |
| **x** = **Ay** | `AjkBk` | `matrixVectorMultiply` | `matrixVector` |
| **x** = **x** + **Ay** | `plusAjkBk` | `matrixVectorMultiply` | `plutMatrixVector` |
| **x** = **A**^T**y** | `AkjBk` | `matrixTVectorMultiply` | `matrixTransposeVector` |


Symmetric matrix operations
Symmetric matrices: **S, Q**
| Operation | Einstein notation | current |
| - | - | - | 
| **x** = **Sy** | `symAjkBk` | `symmetricMatrixVector` |
| **x** = **x** + **Sy** | `plusSymAjkBk` |  `plusSymmetricMatrixVector` |
| **A** = **SB^T** | `symAikBjk` | `symmetricMatrixMatrixTranspose` |
| **S** = **AQA^T** | `AikSymBklAjl` | `matrixSymmetricMatrixMatrixTranspose` |

In the current naming scheme methods like `add`, `scale` and `copy` are overloaded for both matrices and vectors. This works because to add two vectors you'd call `add< M >` and to add two matrices you'd call `add< M, N >`. We could actually do something similar for the symmetric matrix vector multiplication `matrixVector< M, N >` for a non symmetric matrix and `matrixVector< M >` for a symmetric matrix (substitute your preferred naming scheme). I would vote to give symmetric matrix operations a different name, but I don't think it matters too much.

Thoughts?
@castelletto1 @rrsettgast @joshua-white @klevzoff

rrsettgast:
@corbett5 I am wondering if we can test the performance of a general set of functions rather than explicitly unrolling the contractions? In other works instead of having a 2 and a 3 length specialization, we could just have the compiler unroll the loops for us....and it might be just as fast. Other than the indexing of the symmetric tensors, do you see any issues with trying this?

rrsettgast:
> Here are some naming conventions we could use.
> vectors: **x, y, z**
> matrices: **A, B, C**
> 
> Operation	Einstein notation	BlasLapackLA	current
> max absolute entry	?	`vectorNormInf`	`maxAbsolueEntry`
> **x**_i = **y**_i **z**_i	?	?	`elementWiseMultiplication`
> **x** = **y** + **z**	?	`vectorVectorAdd`	`add`
> **A** = **B** + **C**	?	`matrixMatrixAdd`	`add`
> result = **xy**^T	`AiBi`	`vectorDot`	`innerProduct`
> **A** = **x**^T**y**	`AiBj`	?	`outerProduct`
> **x** = **Ay**	`AjkBk`	`matrixVectorMultiply`	`matrixVector`
> **x** = **x** + **Ay**	`plusAjkBk`	`matrixVectorMultiply`	`plutMatrixVector`
> **x** = **A**^T**y**	`AkjBk`	`matrixTVectorMultiply`	`matrixTransposeVector`
> Symmetric matrix operations
> Symmetric matrices: **S, Q**
> 
> Operation	Einstein notation	current
> **x** = **Sy**	`symAjkBk`	`symmetricMatrixVector`
> **x** = **x** + **Sy**	`plusSymAjkBk`	`plusSymmetricMatrixVector`
> **A** = **SB^T**	`symAikBjk`	`symmetricMatrixMatrixTranspose`
> **S** = **AQA^T**	`AikSymBklAjl`	`matrixSymmetricMatrixMatrixTranspose`
> In the current naming scheme methods like `add`, `scale` and `copy` are overloaded for both matrices and vectors. This works because to add two vectors you'd call `add< M >` and to add two matrices you'd call `add< M, N >`. We could actually do something similar for the symmetric matrix vector multiplication `matrixVector< M, N >` for a non symmetric matrix and `matrixVector< M >` for a symmetric matrix (substitute your preferred naming scheme). I would vote to give symmetric matrix operations a different name, but I don't think it matters too much.
> 
> Thoughts?
> @castelletto1 @rrsettgast @joshua-white @klevzoff

Go with the index notation, with different names for symmetric tensors.

corbett5:
> 
> 
> @corbett5 I am wondering if we can test the performance of a general set of functions rather than explicitly unrolling the contractions? In other works instead of having a 2 and a 3 length specialization, we could just have the compiler unroll the loops for us....and it might be just as fast. Other than the indexing of the symmetric tensors, do you see any issues with trying this? We could use a map to translate the symmetric indices.

I can test it out but I'm willing to bet it will be a lot slower, especially with CUDA where to unroll a loop the bounds need to be known at compile time.

rrsettgast:
They are known at compile time.

corbett5:
You wouldn't be able to know until runtime if the operation involves a 2x2 symmetric tensor or a 3x3 symmetric tensor.

rrsettgast:
> You wouldn't be able to know until runtime if the operation involves a 2x2 symmetric tensor or a 3x3 symmetric tensor. Also I'm pretty sure `constexpr int iSize = sizeHelper(srcSymMatrix,0)` would not compile when `srcSymMatrix` is an `ArraySlice`.

Arg. You are correct. You could always pass the dimensions in as template arguments, which could be deduced in the case of c-arrays. Those fixed dimension LvArrays would be nice.

src/tensorOps.hpp
```diff
@@ -22,126 +22,1990 @@
...
+
+  /**
+   * @brief @return Return the determinant of the matrix @p matrix.
+   * @tparam MATRIX The type of @p matrix.
+   * @param matrix The matrix to get the determinant of.
+   * @note This is not implemented for generic sized matrices and calling this is a compilation error.
+   */
+  template< typename MATRIX >
+  LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK inline
+  static int determinant( MATRIX const & matrix )
```
rrsettgast, line 211: @castelletto1 I think it is kind of standard to just use `det()` for the function name....but I rather like spelling it out. Objections?

src/tensorOps.hpp
```diff
@@ -22,126 +22,1990 @@
...
+   * @tparam DST_MATRIX The type of @p dstMatrix.
+   * @tparam SRC_MATRIX The type of @p srcMatrix.
+   * @param dstMatrix The matrix to write the inverse to.
+   * @param srcMatrix The matrix to take the inverse of.
+   * @return The determinant.
+   * @note This is not implemented for generic sized matrices and calling this is a compilation error.
+   */
+  template< typename DST_MATRIX, typename SRC_MATRIX >
+  LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK inline
+  static int invert( DST_MATRIX && LVARRAY_RESTRICT_REF dstMatrix,
```
rrsettgast, line 229:
```suggestion
  static int inverse( DST_MATRIX && LVARRAY_RESTRICT_REF dstMatrix,
```

src/tensorOps.hpp
```diff
@@ -22,126 +22,1990 @@
...
+ * @brief Multiply the entries of @p vector by @p scale.
+ * @tparam M The size of @p vector.
+ * @tparam VECTOR The type of @p vector.
+ * @param vector The vector to scale, of length M.
+ * @param scale The value to scale the entries of @p vector by.
+ * @details Performs the operation @code vector[ i ] *= scale @endcode
+ */
+template< std::ptrdiff_t M, typename VECTOR >
+LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK inline
+void scale( VECTOR && vector, std::remove_reference_t< decltype( vector[ 0 ] ) > const scale )
```
rrsettgast, line 990: If you had a `sizeHelper` that returned the total size of the vector/matrix and a helper to extract the pointer to the beginning of the data, there are a bunch of vector/matrix implementations that could be a single function....which I would prefer on paper.

scale, add, scaledCopy, scaledAdd, elementWiseMultiplication...

src/streamIO.hpp
```diff
@@ -310,7 +310,7 @@ static void stringToArray( Array< T, NDIM, PERMUTATION, INDEX_TYPE, DATA_VECTOR_
  */
 template< typename T, int NDIM, int USD, typename INDEX_TYPE >
 std::ostream & operator<<( std::ostream & stream,
-                           ArraySlice< T const, NDIM, USD, INDEX_TYPE > const & slice )
+                           ArraySlice< T, NDIM, USD, INDEX_TYPE > const & slice )
```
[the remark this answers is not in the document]
corbett5: It doesn't require `T` to be non const. But if you force `T const` then to print a non const slice you need to to `std::cout << slice.toConst();`

src/tensorOps.hpp
corbett5: Slice's in general don't have contiguous storage and you need the double loop to do `x[ i ][ j ]`.

PR #164, opened 2020-06-17
Addresses some Tribol requests.

corbett5:
- Now uses vanilla Chai 2.1.1, and can build without.
- Compiles with all four combinations of `USE_OPENMP` and `USE_CUDA`.
- Can specify the path to `RAJA` and `CHAI` to build against.
- Every class is now templated on the buffer type.
- Removed all default template arguments.
- Removed the `NDIM` template parameter of `Array`.
- Replaced `push_back` with `emplace_back`.
- Replaced single valued `insert` with `emplace`.
- All multi valued append/insertion methods take iterators instead of a pointer and a size.
- Added `begin` and `end` to `ArraySlice` and removed `ArrayOfArrayView::getIterableArray`.
- All of the unit tests use `TYPED_TEST` and there are no longer any tests that are only run with CUDA.


src/Array.hpp
```diff
@@ -64,40 +59,40 @@ namespace LvArray
  *         one template argument that describes the type of the data being stored (T).
  */
 template< typename T,
-          int NDIM,
-          typename PERMUTATION=camp::make_idx_seq_t< NDIM >,
-          typename INDEX_TYPE=std::ptrdiff_t,
-          template< typename > class BUFFER_TYPE=NewChaiBuffer >
+          typename PERMUTATION,
+          typename INDEX_TYPE,
+          template< typename > class BUFFER_TYPE >
```
corbett5, line 64: So I removed the `NDIM` template argument because the permutation carries the dimension as well and by removing the default template arguments you'll always have to specify the permutation anyways. I'm now having second thoughts about doing this for the following reasons

- Both `ArrayView` and `ArraySlice` have `NDIM` as the second template parameter and this breaks the symmetry.
- If you do the following
    ```c++
    template< typename T, typename PERMUTATION=camp::make_idx_seq_t< 2 > >
    using array2d = Array< T, PERMUTATION, std::ptrdiff_t, NewChaiBuffer >
    ```
    the following is valid `array2d< int, RAJA::PERM_I > really_a_1d_array;`
- It makes overloading on dimension harder
    ```c++
    // Operate on a 1D array.
    void operate( Array< int, 1, RAJA::PERM_I, std::ptrdiff_t, NewChaiBuffer > const & x );

    // Operate on a 2D array.
    template< typename PERM >
    void operate( Array< int, 2, PERM, std::ptrdiff_t, NewChaiBuffer > const & x );
    ```
    becomes
    ```c++
    // Operate on a 1D array.
    template< typename PERM >
    std::enable_if_t< getDimension( PERM ) == 1 >
    operate( Array< int, RAJA::PERM_I, std::ptrdiff_t, NewChaiBuffer > const & x );

    // Operate on a 2D array.
    template< typename PERM >
    std::enable_if_t< getDimension( PERM ) == 2 >
    void operate( Array< int, RAJA::PERM_I, std::ptrdiff_t, NewChaiBuffer > const & x );
    ```
@rrsettgast  thoughts?

rrsettgast: I think keeping `NDIM` in the parameter list makes it much clearer what is happening....for the reasons that you list, and probably some other ones we haven't encountered yet. The redundancy isn't the worst thing in the world. Does `NDIM` help to check that you specified an appropriate permutation?

corbett5: Yes, with `NDIM` in the template list `Array< int, 2, RAJA::PERM_I >` would fail with a static_assert. I'll add it back in.

corbett5: Done.

src/CRSMatrix.hpp
```diff
@@ -138,7 +143,7 @@ class CRSMatrix : protected CRSMatrixView< T, COL_TYPE, INDEX_TYPE >
    * @param src the SparsityPattern to convert.
    */
   inline
-  void stealFrom( SparsityPattern< COL_TYPE, INDEX_TYPE > && src )
+  void stealFrom( SparsityPattern< COL_TYPE, INDEX_TYPE, BUFFER_TYPE > && src )
```
rrsettgast, line 146:
```suggestion
  void assimilate( SparsityPattern< COL_TYPE, INDEX_TYPE, BUFFER_TYPE > && src )
```

src/CRSMatrix.hpp
```diff
@@ -151,7 +156,7 @@ class CRSMatrix : protected CRSMatrixView< T, COL_TYPE, INDEX_TYPE >
     // Reallocate to the appropriate length
     bufferManipulation::reserve( m_entries, 0, src.nonZeroCapacity() );
 
-    ParentClass::stealFrom( reinterpret_cast< SparsityPatternView< COL_TYPE, INDEX_TYPE > && >( src ) );
+    ParentClass::stealFrom( reinterpret_cast< SparsityPatternView< COL_TYPE, INDEX_TYPE, BUFFER_TYPE > && >( src ) );
```
rrsettgast, line 159:
```suggestion
    ParentClass::assimilate( reinterpret_cast< SparsityPatternView< COL_TYPE, INDEX_TYPE, BUFFER_TYPE > && >( src ) );
```

src/CRSMatrix.hpp
```diff
@@ -169,31 +174,35 @@ class CRSMatrix : protected CRSMatrixView< T, COL_TYPE, INDEX_TYPE >
    * @brief @return A reference to *this reinterpreted as a CRSMatrixView< T, COL_TYPE, INDEX_TYPE const >.
    */
   constexpr inline
-  CRSMatrixView< T, COL_TYPE, INDEX_TYPE const > const & toView() const LVARRAY_RESTRICT_THIS
-  { return reinterpret_cast< CRSMatrixView< T, COL_TYPE, INDEX_TYPE const > const & >( *this ); }
+  CRSMatrixView< T, COL_TYPE, INDEX_TYPE const, BUFFER_TYPE > const &
+  toView() const LVARRAY_RESTRICT_THIS
+  { return reinterpret_cast< CRSMatrixView< T, COL_TYPE, INDEX_TYPE const, BUFFER_TYPE > const & >( *this ); }
```
rrsettgast, line 179: We do need to get rid of the `reinterpret_cast` at some point. This would be replaced by the construction of a view, and the return of that view rather than a reference. Can this be done without downstream changes?

src/NewChaiBuffer.hpp
```diff
@@ -49,6 +48,38 @@ inline chai::ArrayManager & getArrayManager()
...
 
+inline chai::ExecutionSpace toChaiExecutionSpace( MemorySpace const space )
+{
+  if( space == MemorySpace::NONE )
+    return chai::NONE;
+  if( space == MemorySpace::CPU )
+    return chai::CPU;
+#if defined(USE_CUDA)
+  if( space == MemorySpace::GPU )
+    return chai::GPU;
```
rrsettgast, line 59: Do we even want to use the chai execution spaces? It seems like they should just alias the `MemorySpace`

src/bufferManipulation.hpp
```diff
@@ -16,32 +16,61 @@
...
 #include <utility>
 
 namespace LvArray
 {
+
+/**
+ * @enum MemorySpace
+ * @brief An enum containing the available memory spaces.
+ */
+enum class MemorySpace
```
rrsettgast, line 37: Does umpire have a similar enum?

src/streamIO.hpp
```diff
@@ -354,8 +353,9 @@ std::ostream & operator<<( std::ostream & stream,
  * @param view The SortedArrayView to output.
  * @return @p stream .
  */
-template< typename T, typename INDEX_TYPE >
-std::ostream & operator<< ( std::ostream & stream, SortedArrayView< T const, INDEX_TYPE > const & view )
+template< typename T, typename INDEX_TYPE, template< typename > class BUFFER_TYPE >
+std::ostream & operator<< ( std::ostream & stream,
```
rrsettgast, line 357: in addition to these stream operators, it would be nice to have an output option that would dump using `printf` so that we could use it on device. For instance a `printRow( INDEX_TYPE const row )`.

src/templateHelpers.hpp
```diff
@@ -132,38 +136,24 @@ void forEachArg( F && f, ARG && arg, ARGS && ... args )
...
- * @brief A helper struct since host variables can't be accessed in device code.
- */
-template< bool ... ARGS >
-struct Conjunction
-{
-  /// Evaluates to true of all of ARGS are true.
-  static constexpr bool value = conjunction< ARGS ... >;
-};
+template< typename ... TYPES >
+using all_of_t = camp::concepts::metalib::all_of_t< TYPES ... >;
```
rrsettgast, line 150: nice

benchmarks/benchmarkArray1DR2TensorMultiplicationKernels.hpp
```diff
@@ -220,7 +220,7 @@ class ArrayOfR2TensorsRAJA : private ArrayOfR2TensorsNative< PERMUTATION >
   }
 
   ~ArrayOfR2TensorsRAJA()
-  { this->m_c.move( chai::CPU ); }
+  { this->m_c.move( MemorySpace::CPU ); }
```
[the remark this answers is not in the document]
corbett5: It's an enum I defined myself. I think camp and umpire both have similar features but this works fine for now.

cmake/SetupTPL.cmake
```diff
@@ -32,6 +25,58 @@ set(ENABLE_RAJA ON CACHE BOOL "")
...
+    if(NOT ENABLE_UMPIRE)
+        message(FATAL_ERROR "umpire must be enabled to use chai.")
+    endif()
+
+    if(NOT ENABLE_RAJA)
+        message(FATAL_ERROR "RAJA must be enabled to use chai.")
+    endif()
+
+    if(NOT EXISTS ${CHAI_DIR})
+        set(CHAI_DIR ${GEOSX_TPL_DIR}/chai)
```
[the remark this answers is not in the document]
corbett5: The patch in the tpl repo needs to be applied until the next release. But building off of the main release and not my `feature/corbett/changes-for-geosx` branch.

src/Array.hpp
```diff
@@ -236,35 +235,25 @@ class Array : public ArrayView< T,
...
    * @tparam DIMS Variadic list of integral types.
    * @param newDims The new dimensions, must be of length NDIM.
-   * @note This does not preserve the values in the Array unless the default permutation is used.
+   * @note This does not preserve the values in the Array unless NDIM == 1.
+   * @return void.
    */
   template< typename ... DIMS >
   LVARRAY_HOST_DEVICE
-  void resize( DIMS... newDims )
+  std::enable_if_t< sizeof ... ( DIMS ) == NDIM && all_of_t< std::is_integral< DIMS > ... >::value >
```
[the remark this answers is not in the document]
corbett5: void
``` c++
template< bool CONDITION, typename RETURN=void >
using enable_if_t = typename std::enable_if< CONDITION, RETURN >;
```

src/NewChaiBuffer.hpp
corbett5: The Chai api needs a `chai::ExecutionSpace` so we need to have a method to convert between the two.

benchmarks/benchmarkArray1DR2TensorMultiplicationKernels.cpp
```diff
@@ -64,10 +64,10 @@ namespace benchmarking
 
 
 template< typename VALUE_TYPE_CONST, int USD >
-RAJA_INLINE LVARRAY_HOST_DEVICE constexpr
-void R2TensorMultiply( LvArray::ArraySlice< VALUE_TYPE_CONST, 2, USD > const & a,
-                       LvArray::ArraySlice< VALUE_TYPE_CONST, 2, USD > const & b,
-                       LvArray::ArraySlice< VALUE_TYPE, 2, USD > const & c )
+inline LVARRAY_HOST_DEVICE constexpr
```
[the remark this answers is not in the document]
corbett5: Nope.

src/Array.hpp
```diff
@@ -385,91 +379,61 @@ class Array : public ArrayView< T,
...
+   * @param args A variadic pack of values to construct the new value from.
    * @note This method is only available on 1D arrays.
    * @return void.
    */
-  template< int N=NDIM >
-  std::enable_if_t< N == 1 >
-  push_back( T const & value )
+  template< typename ... ARGS, int _NDIM=NDIM >
+  std::enable_if_t< _NDIM == 1 >
+  emplace_back( ARGS && ... args )
```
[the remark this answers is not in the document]
corbett5: `emplace_back` is a drop in replacement for `push_back` and it's supported by `std::vector` as well. I would prefer not to also have `push_back` because it would add two methods `push_back( T const &)` and `push_back( T && )` that add no new functionality.

src/CRSMatrix.hpp
corbett5: Yes everything should compile since we currently have
```c++
CRSMatrixView< T > const & CRSMatrix< T >::toView() const
{ return reinterpret_cast<CRSMatrixView< T > const & >( *this ); }
...
CRSMatrix< T > matrix;
CRSMatrixView< T > const & view = matrix.toView();
```
and if we change `toView() to return by value it will still work
```c++
CRSMatrixView< T > CRSMatrix< T >::toView() const
{ return CRSMatrixView< T >( m_entries, m_values, m_sizes, m_offsets ); }
...
CRSMatrix< T > matrix;
CRSMatrixView< T > const & view = matrix.toView();
```
However I think best practices dictate that now that `toView` returns by value we they type of `view` should be `CRSMatrixView< T > const` and not `CRSMatrixView< T > const &`.

Furthermore there could be runtime issues if the user writes code that assumes the view and the owner are the same object. Currently this would pass
```c++
Array1d< int > array;
ArrayView1d< int > const & view = array.toView();
array.resize( 1 );
EXPECT_EQ( view.size(), 1 );
EXPECT_NE( view.data(), nullptr );
```
However once we switch to returning by value this will fail.

src/bufferManipulation.hpp
corbett5: Probably, I think camp does too.

src/streamIO.hpp
corbett5: #165

src/ArrayOfArrays.hpp
```diff
@@ -37,13 +37,26 @@ class ArrayOfSets;
...
+          typename INDEX_TYPE,
+          template< typename > class BUFFER_TYPE >
+class ArrayOfArrays : protected ArrayOfArraysView< T, INDEX_TYPE, false, BUFFER_TYPE >
 {
+  /// An alias for the parent class.
+  using ParentClass = ArrayOfArraysView< T, INDEX_TYPE, false, BUFFER_TYPE >;
+
 public:
+  /// An alias for the type contained in the arrays.
+  using value_type = T;
```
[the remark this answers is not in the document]
corbett5: Done.

src/ArrayOfArrays.hpp
```diff
@@ -124,7 +136,7 @@ class ArrayOfArrays : protected ArrayOfArraysView< T, INDEX_TYPE >
    * @param src the ArrayOfSets to convert.
    */
   inline
-  void stealFrom( ArrayOfSets< T, INDEX_TYPE > && src )
+  void stealFrom( ArrayOfSets< T, INDEX_TYPE, BUFFER_TYPE > && src )
```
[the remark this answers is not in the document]
corbett5: Done.

PR #166, opened 2020-06-19
Make arrays available in Python as np.ndarray objects

jameshcorbett:
Adds functions for returning Numpy views of ArrayView and SortedArrayView objects. With a Python extension module, these views can then be manipulated in Python. All that's needed is to pass the objects to one of the functions added in `src/python`. These will return a `PyObject_t *` which should be passed to the Python interpreter. There are two python extension modules, used for testing, in `unitTests/python`.

### DIFF
--- a/cmake/SetupTPL.cmake
+++ b/cmake/SetupTPL.cmake
@@ -120,6 +120,27 @@ if(ENABLE_CALIPER)
     set(thirdPartyLibs ${thirdPartyLibs} caliper)
 endif()
 
+################################
+# Python
+################################
+if(ENABLE_PYTHON)
+    include(cmake/FindPython.cmake)
+    message(STATUS "Using Python Include: ${PYTHON_INCLUDE_DIRS}")
+    # include_directories(${PYTHON_INCLUDE_DIRS})
+    # # if we don't find python, throw a fatal error
+    # if(NOT PYTHON_FOUND)
+    #     message(FATAL_ERROR "ENABLE_PYTHON is true, but Python wasn't found.")
+    # endif()
+
+    # include(cmake/thirdparty/FindNumPy.cmake)
+    # message(STATUS "Using NumPy Include: ${NUMPY_INCLUDE_DIRS}")
+    # include_directories(${NUMPY_INCLUDE_DIRS})
+    # # if we don't find numpy, throw a fatal error
+    # if(NOT NUMPY_FOUND)
+    #     message(FATAL_ERROR "ENABLE_PYTHON is true, but NumPy wasn't found.")
+    # endif()
+endif()
+
 
 
 
--- a/cmake/SetupTPL.cmake
+++ b/cmake/SetupTPL.cmake
@@ -126,19 +126,6 @@ endif()
 if(ENABLE_PYTHON)
     include(cmake/FindPython.cmake)
     message(STATUS "Using Python Include: ${PYTHON_INCLUDE_DIRS}")
-    # include_directories(${PYTHON_INCLUDE_DIRS})
-    # # if we don't find python, throw a fatal error
-    # if(NOT PYTHON_FOUND)
-    #     message(FATAL_ERROR "ENABLE_PYTHON is true, but Python wasn't found.")
-    # endif()
-
-    # include(cmake/thirdparty/FindNumPy.cmake)
-    # message(STATUS "Using NumPy Include: ${NUMPY_INCLUDE_DIRS}")
-    # include_directories(${NUMPY_INCLUDE_DIRS})
-    # # if we don't find numpy, throw a fatal error
-    # if(NOT NUMPY_FOUND)
-    #     message(FATAL_ERROR "ENABLE_PYTHON is true, but NumPy wasn't found.")
-    # endif()
 endif()
 
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -45,10 +45,10 @@ set(lvarray_sources
     )
 
 if(ENABLE_PYTHON)
-    set(lvarray_headers
-        python/numpyConversion.hpp
-        python/numpyArrayView.hpp
-        python/numpySortedArrayView.hpp
+    list( APPEND lvarray_headers
+          python/numpyConversion.hpp
+          python/numpyArrayView.hpp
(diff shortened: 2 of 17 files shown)


cmake/SetupTPL.cmake
```diff
@@ -120,6 +120,27 @@ if(ENABLE_CALIPER)
...
+    #     message(FATAL_ERROR "ENABLE_PYTHON is true, but Python wasn't found.")
+    # endif()
+
+    # include(cmake/thirdparty/FindNumPy.cmake)
+    # message(STATUS "Using NumPy Include: ${NUMPY_INCLUDE_DIRS}")
+    # include_directories(${NUMPY_INCLUDE_DIRS})
+    # # if we don't find numpy, throw a fatal error
+    # if(NOT NUMPY_FOUND)
+    #     message(FATAL_ERROR "ENABLE_PYTHON is true, but NumPy wasn't found.")
+    # endif()
```
corbett5, line 141: Delete.

src/python/numpyArrayView.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace LvArray
+{
+
+namespace python
+{
+
+/**
+ * Return a Numpy view of an Array
+ * @param arr the ArrayView to convert to numpy.
```
corbett5, line 35: Document the template parameters with `@tparam` and add a `@brief` describing what you can and can't do with the returned array. For example you can't resize it and if `T` is `const` you can't assign to it either.

src/python/numpyConversion.hpp
```diff
@@ -0,0 +1,119 @@
...
+#define PY_ARRAY_UNIQUE_SYMBOL LvArray_ARRAY_API
+
+#include <Python.h>
+#include <numpy/arrayobject.h>
+
+#include "../StringUtilities.hpp"
+#include "../IntegerConversion.hpp"
+
+#include <typeindex>
+#include <type_traits>
```
corbett5, line 36: Add a comment about how `<Python.h>` has to be included before everything else and add a `// Source includes`` and `// System includes` comment.

src/python/numpyConversion.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace internal
+{
+
+/**
+ * @brief Get the Numpy type-flag corresponding to a `std::type_info` object.
+ * @param tid the type_info object to use
+ */
+int getFlags( const std::type_info& tid ){
+    if ( std::type_index( tid ) == std::type_index( typeid( char ) ) )
```
corbett5, line 52: Can you add unsigned variations on all of these if numpy supports it? Be sure to have both `signed` and `unsigned char` since the signedness of `char` is implementation defined.

src/python/numpySortedArrayView.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Return a Numpy view of a SortedArray
+ * @param arr the SortedArrayView to convert to numpy.
+ */
+template< typename T, typename INDEX_TYPE, template<typename> class BUFFER_TYPE >
+PyObject * create( SortedArrayView< T const, INDEX_TYPE, BUFFER_TYPE > const & arr ){
+    arr.move( MemorySpace::CPU );
+    INDEX_TYPE dims = arr.size();
+    INDEX_TYPE strides = 1;
```
corbett5, line 41: `INDEX_TYPE const`.

src/CMakeLists.txt
```diff
@@ -45,6 +44,13 @@ set(lvarray_sources
     totalview/tv_data_display.c 
     )
 
+if(ENABLE_PYTHON)
+    set(lvarray_headers
+        python/numpyConversion.hpp
+        python/numpyArrayView.hpp
+        python/numpySortedArrayView.hpp
+        )
```
rrsettgast, line 52: Shouldn't this be
```suggestion
    list( APPEND lvarray_headers
          python/numpyConversion.hpp
          python/numpyArrayView.hpp
          python/numpySortedArrayView.hpp
          )
```

?

src/python/numpyArrayView.hpp
rrsettgast: What would happen if the caller resizes the array? It seems like the original `Array` would be invalid. Can we add a validity check for the variables that were provided to python upon calling the c-code again?

corbett5: I'm not sure what you mean.

NumPy prevents you from resizing externally allocated arrays so we're safe if someone in python tries to resize a NumPy array they got from geosx.

And we agreed that it is invalid to keep references to objects obtained from geosx around in-between time steps. There may be a way to keep track of all the objects that geosx exports into python and invalidate them each time it does a time step but it wouldn't be pretty.

rrsettgast: Well is you cannot resize an externally allocated array in numpy, then it is a pointless request. I read your comment incorrectly as a "shouldn't resize" instead of "can't resize". 

So what can you and can't you do on the python side? 
You CANNOT reallocate. 
Can you modify the data even if the source is `ArrayView< T const >`?

corbett5: You cannot reallocate and if `T` is const you can't modify the values.

rrsettgast: ok. I read the intent of your entire request incorrectly.

jameshcorbett: You get reasonably nice errors:

```
62: Traceback (most recent call last):
62:   File "/g/g12/corbett8/LvArray/unitTests/python/testPythonSortedArrayDriver.py", line 47, in test_resizing
62:     array_from_c.resize((100,))
62: ValueError: cannot resize this array: it does not own its data
```

```
62: Traceback (most recent call last):
62:   File "/g/g12/corbett8/LvArray/unitTests/python/testPythonSortedArrayDriver.py", line 38, in test_modification
62:     array_from_c *= 2
62: ValueError: output array is read-only
```

jameshcorbett:
I pushed some commits (to be squashed later) which I think address all the comments so far.

corbett5:
This branch is still from your fork. I recommend we close this and switch over to `feature/corbetts/python` (#186). We can then both add commits to that going forward.

Issue #160, opened 2020-06-10
Tribol use case issues


rrsettgast:
These comments/issues were provided by steve glenn after some effort to integrate LvArray into Tribol.

1.	git/config contains references to ssh://github.com/LLNL instead of https://github.com/LLNL, which causes failures if ssh is not configured.
- [x] Fixed

2.	Docs fail to build (hang) on Lassen.  
- [x] Fixed

3.	Doc build cannot be disabled using ENABLE_DOCS=Off
- [x] Fixed

4.	Library and header exports are missing from install directory, so clients are forced to pull from the build directory.  The install step emits a “Cannot create regular file” error.
- [ ] Fixed

5.	The build should complete for all four combinations of ENABLE_CUDA=On/Off and ENABLE_OPENMP=On/Off.
- [x] Fixed

6.	LvArray cannot be built with xlc V16.1.1 on Lassen.  According to LC documentation, https://hpc.llnl.gov/training/tutorials/using-lcs-sierra-system#Compilers, “… recommended and supported compilers are those delivered from IBM (XL and Clang ) and NVIDIA (NVCC).”  A fix from IBM may also be needed for xlc to work with LvArray.
- [ ] Fixed

7.	LvArray should be buildable against independent builds of recent versions of Axom, CHAI/Umpire and RAJA, not only those in /usr/gapps/GEOSX/thirdPartyLibs.  This is important since applications may need to customize various options.  
For example, Tribol needs Slic, Spin, Quest, and Primal from Axom, but /usr/gapps/GEOSX/thirdPartyLibs/2020-05-23/install-lassen-clang@upstream-release only contains slic and lumberjack.
- [x] Fixed. Hard dependency on RAJA, soft dependency on Chai and Caliper. No dependency on Axom.

8.	The documentation needs better examples demonstrating use of two-dimensional Arrays, ArraySlices, and ArrayOfArrays.  Currently, the unit tests are the only available examples.
#159 

9.	Consider adding support for emplace_back().  In addition to making LvArray conform more closely to the std::vector interface, some applications may want this instead of push_back() for better performance.
- [x] Fixed.

PR #185, opened 2020-07-21
Docs/corbett/lots o sphinx

### LINKED ISSUE #159: Examples with simple use cases.
The unit tests are currently kind of difficult to read for newcomers. We need to generate a set of examples that illustrate common usage for each of the classes. Also we should document the examples in sphinx. The sphinx is probably sufficient, but it would be nice to have the usage examples coded up and run with the unit tests s.t. any changes will be caught. Then we can just include the examples into the sphinx directly.

### LINKED ISSUE #178: Port Eigenvalue/Eigenvector decompositions from R2SymTensor
@andrembcosta  needs the spectral decompositions that used to live in R2SymTensor before it was removed from the code for phase field fracture.

corbett5:
This is pretty un-reviewable but the main thing is I'd like you to read over the documentation. Is there an easy way to set up RTD to build this branch? In addition to the completely new sphinx docs this PR did the following

- Adds eigenvalue and eigenvector solvers for 2x2 and 3x3 symmetric matrices.
- Adds a few new symmetric matrix methods such as `symTrace`.
- Adds some portable math functions in `math.hpp`.
- Conglomerates some files and adds namespaces. For example `SetFPE`, `StringUtilities` and `stackTrace` are now all one file `system` in the `LvArray::system` namespace. The reason for the namespaces is that it makes the doxygen much easier to navigate.
- Adds doxygen groupings for class methods.
- Adds an eigenvalue and eigenvector benchmark.
- Adds two `ArrayOfArrays` benchmarks.

Closes #159
Closes #178
Addresses #160


rhornung67:
@corbett5 you can activate the doc build for this branch in RTD. Click on versions, find it in the list, click on it, and activate it. Docs will be built automatically on nest commit to branch, or you can build it manually on RTD.

corbett5:
Okay the docs are up at
https://lvarray.readthedocs.io/en/docs-corbett-lots-o-sphinx/
The doxygen links are broken though. @rrsettgast do you know how to properly link to the doxygen on RTD? I have the doxygen directory as `../../doxygen/html/` which works fine for local builds but apparently not for RTD. In geosx the doxygen link is `../../doxygen_output/html/`.

corbett5:
Okay the doxygen links now work on RTD. I broke them for local builds though. The issue is that locally I link to doxygen via `../../doxygen/html/file.html` since the sphinx html files are in `build_dir/docs/sphinx/html` and the doxygen html files are in `build_dir/docs/doxygen/html`. However on RTD I need to link with just `doxygen/html/file.html` which implies that the doxygen files are built inside the sphinx html file directory.

Is there some way to fix this that doesn't involve duplicating the RTD directory structure for local builds? Does sphinx have variables you can set and use in links?

PR #186, opened 2020-07-22
Python interface

corbett5:


src/python/PyFunc.hpp
```diff
@@ -0,0 +1,95 @@
...
+{
+public:
+
+  PythonFunction( PyObject * pyfunc ):
+    m_function( pyfunc )
+  { internal::xincref( pyfunc ); }
+
+  void operator()( ARGS ... args )
+  {
+    // if an error is already set, don't invoke the callback
```
corbett5, line 57: @jameshcorbett what is with this comment? It appears to me that the function is called if an error is set.

jameshcorbett: I think we were arguing about what to do if the callback raises an error. If we just ignore it and try to invoke the callback again (since it may be invoked multiple times before returning control to the user's script) Python raises a `SystemError` which is not pretty and the actual exception is lost. Originally I suggested that once the callback raises, we just stop invoking it but continue as usual, and let the callback's exception (eventually) propagate to the user once the solver had finished. You didn't like that idea very much, for good reason. So we settled on raising a C++ exception (`throw PythonError();`) instead to return control to the user's script as soon as the callback raises. So the comment isn't entirely wrong in that you don't invoke the callback again once it has raised an error, but I think it should be reworded and moved down to the `throw` line.

jameshcorbett: The usual thing to do would just be to propagate `nullptr` all the way back down to the Python C interface, but that wasn't an option for the physics solver.

corbett5: I remember this now.

That sounds like a good idea.

corbett5:
LGTM

jameshcorbett (approved):

PR #188, opened 2020-07-28
Changes for icpc.

corbett5:


rrsettgast:
@corbett5 What is the error message for this issue?

corbett5:
It varies, but the issue is that a `constexpr` member function is implicitly `const` with intel. So for example you can't make assignment operators `constexpr` because they assign to the members of the class and can't be `const`.

src/ArrayView.hpp
```diff
@@ -189,7 +189,7 @@ class ArrayView
    * anotherView = std::move( view );
    * @endcode
    */
-  inline LVARRAY_HOST_DEVICE constexpr
+  inline LVARRAY_HOST_DEVICE LVARRAY_INTEL_CONSTEXPR
```
rrsettgast, line 192: try module load a newer gcc instead of this?

PR #192, opened 2020-08-22
Bugfix/corbett/benchmarks

corbett5:
- Fixes some bugs in the benchmarks.
- Removes `reinterpret_cast` except from `Array` and `ArrayView`.
- Changes for the XL compiler, some of them are kind of hacky and I think may be due to compiler bugs.
- Added a floating point exception guard.


corbett5:
Closes https://github.com/GEOSX/GEOSX/issues/1117

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "src/cmake/blt"]
 	path = cmake/blt
-	url = git@github.com:LLNL/blt.git
+	url = https://github.com/LLNL/blt.git
```
rrsettgast, line 3: I think I had to do this already.

benchmarks/benchmarkMatrixMatrix.cpp
```diff
@@ -117,20 +117,20 @@ void pointerRAJA( benchmark::State & state )
...
+INDEX_TYPE const OMP_M = 4 * SERIAL_M;
 #endif
 
 #if defined(LVARRAY_USE_CUDA) && defined(LVARRAY_USE_CHAI)
-INDEX_TYPE const CUDA_N = SERIAL_N;
-INDEX_TYPE const CUDA_L = SERIAL_L;
-INDEX_TYPE const CUDA_M = SERIAL_M;
+INDEX_TYPE const CUDA_N = 16 * SERIAL_N;
+INDEX_TYPE const CUDA_L = 16 * SERIAL_L;
+INDEX_TYPE const CUDA_M = 16 * SERIAL_M;
```
rrsettgast, line 133: What is/was happening here that required a change?

src/SortedArrayView.hpp
```diff
@@ -86,28 +86,38 @@ class SortedArrayView
    * @brief Default move constructor, performs a shallow copy.
    * @param src The SortedArray to be moved from.
    */
-  LVARRAY_HOST_DEVICE inline
+  LVARRAY_HOST_DEVICE constexpr inline
```
rrsettgast, line 89: Is there a reason we would want to have a `LVARRAY_CONSTEXPR`?

src/limits.hpp
```diff
@@ -51,6 +51,35 @@ struct NumericLimits : public std::numeric_limits< T >
   static constexpr T denorm_min = std::numeric_limits< T >::denorm_min();
 };
 
+/**
+ * @struct NumericLimitsNC
```
rrsettgast, line 55: 1. What does the NC stand for?
2. Do we want to have a preprocessor option to just alias the std::numeric_limits members? That way we can swap it out when we move to c++17?

src/ArrayView.hpp
```diff
@@ -303,7 +303,15 @@ class ArrayView
...
   INDEX_TYPE size() const noexcept
-  { return indexing::multiplyAll< NDIM >( m_dims ); }
+  {
+    // Note: This used to be done with a recursive template but XL-release would produce incorrect results.
+    // Specifically in exampleArray it would return an "old" size even after being updated, strange.
+    INDEX_TYPE val = m_dims[ 0 ];
+    for( int i = 1; i < NDIM; ++i )
+    { val *= m_dims[ i ]; }
+
+    return val;
```
rrsettgast, line 313:
```suggestion

#if defined( __ibmxl__ ) && defined( __CUDA_ARCH__ )
    INDEX_TYPE val = m_dims[ 0 ];
    for( int i = 1; i < NDIM; ++i )
    { val *= m_dims[ i ]; }

    return val;
#else
    return indexing::multiplyAll< NDIM >( m_dims );
#endif  
```

unitTests/testTensorOpsEigen.cpp
```diff
@@ -189,12 +189,13 @@ class TestEigendecomposition : public ::testing::Test
 
       // Construct a diagonal matrix of the eigenvalues
       FLOAT lambda[ tensorOps::SYM_SIZE< M > ];
```
rrsettgast, line 191: Shouldn't this be `FLOAT lambda[M]`?

.gitmodules
corbett5: Yeah this will get rebased out.

benchmarks/benchmarkMatrixMatrix.cpp
corbett5: These are the sizes of the matrices multiplied together in the serial, threaded and gpu benchmarks. These aren't important changes.

unitTests/testTensorOpsEigen.cpp
corbett5: It needs to be a symmetric matrix to leverage the `tensorOps` operations.

src/limits.hpp
corbett5: Non-constexpr.

I don't think it's a standards issue. The reason `NumericLimits` exists is so that you can have access to the value returned by the **host** function `std::numeric_limits::epsilon()`. The reason `NumericLimitsNC` exists is because occasionally if you capture a constexpr variable in a lambda and then do something like `EXPECT_LT( x, epsilon )` you'll get an undefined reference at link time to `epsilon`. I don't think this really has a use case outside of testing so I could move it there.

src/SortedArrayView.hpp
corbett5: We don't have an `LVARRAY_CONSTEXPR`, although now we do have an `LVARRAY_INTEL_CONSTEXPR` which is a bad name because it is only `constexpr` if not using the intel compiler.

src/ArrayView.hpp
corbett5: I wanted to minimize the divergence between XL and the other compilers and my guess would be that both are equally fast, but I'd happily accept this change.

src/SortedArrayView.hpp
rrsettgast: I thought it really only applies to a `constexpr` function as well? Perhaps we should have a `LVARRAY_CONSTEXPR` and an `LVARRAY_CONSTEXPR_FUNC`...didn't we used to have that?

src/limits.hpp
rrsettgast: that sounds like a nice option.

src/ArrayView.hpp
rrsettgast: I take the point of view that this is a problem with the XL compiler, so we should have to desired implementation, and the XL implementation.

unitTests/testTensorOpsEigen.cpp
rrsettgast: the lambda's are the eigenvalues...of which there are 3 for a 3x3, whether it is symmetric or not.

corbett5: On line 202 I call
```
tensorOps::AikSymBklAjl< M >( matrices[ i ], Q, lambda );
```
So `lambda` needs to be in symmetric notation for that call to work.

src/SortedArrayView.hpp
corbett5: I think so but why should we define a macro that just expands to `constexpr`? We wouldn't ever think of doing something like `#define LVARRAY_CONST const` so I don't see why `constexpr` should be any different.

corbett5:
Closes https://github.com/GEOSX/LvArray/issues/45

corbett5:
Closes https://github.com/GEOSX/LvArray/issues/160

src/ArrayView.hpp
corbett5: Done.

rrsettgast (approved):

PR #199, opened 2020-09-17
Creating a default set of floating point exceptions

af1990:
Adding a simple function defining the default set of FPE we want to check.


unitTests/CMakeLists.txt
```diff
@@ -87,6 +87,11 @@ blt_add_executable( NAME testFloatingPointExceptions
                     OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
                     DEPENDS_ON gtest lvarray ${lvarray_dependencies} )
 
+# Need to avoid optimization to catch invalid operations
+if( APPLE AND ${CMAKE_CXX_COMPILER} MATCHES "clang" )
```
corbett5, line 91: Ahh is this because APPLE + "clang" doesn't respect the no-inline attribute?

corbett5 (approved):
Thanks!

unitTests/CMakeLists.txt
af1990: I think so ... I am not a Mac expect!

rrsettgast (approved):

castelletto1 (approved):
Looks good to me. Thanks for the fixes!

PR #200, opened 2020-09-22
Feature/rrsettgast/update tensor ops

rrsettgast:
Make the calls to tensor ops a little more explicit by including the RHS with indices. So for instance `AikBjk` becomes `Rij_eq_AikBjk`. 

This is motivated by some difficulties I had with using the ops. Of course we can just recognize that the result is always `Ri` or `Rij`, but it is better imo to be explicit. 

I also changed the name of the template parameters `M,N,P` to `ISIZE, JSIZE, KSIZE` respectively. 

Associated changes in GEOSX are in https://github.com/GEOSX/GEOSX/pull/1170

### DIFF
--- a/host-configs/LLNL/lassen-clang@upstream.cmake
+++ b/host-configs/LLNL/lassen-clang@upstream.cmake
@@ -54,7 +54,7 @@ set(CMAKE_CUDA_FLAGS_RELWITHDEBINFO "-g -lineinfo ${CMAKE_CUDA_FLAGS_RELEASE}" C
 set(CMAKE_CUDA_FLAGS_DEBUG "-g -G -O0 -Xcompiler -O0" CACHE STRING "")
 
 # Uncomment this line to make nvcc output register usage for each kernel.
- set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --resource-usage" CACHE STRING "" FORCE)
+#set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --resource-usage" CACHE STRING "" FORCE)
 
 # GTEST options
 set(ENABLE_GTEST_DEATH_TESTS OFF CACHE BOOL "")
--- a/host-configs/LLNL/lassen-clang@upstream.cmake
+++ b/host-configs/LLNL/lassen-clang@upstream.cmake
@@ -54,7 +54,7 @@ set(CMAKE_CUDA_FLAGS_RELWITHDEBINFO "-g -lineinfo ${CMAKE_CUDA_FLAGS_RELEASE}" C
 set(CMAKE_CUDA_FLAGS_DEBUG "-g -G -O0 -Xcompiler -O0" CACHE STRING "")
 
 # Uncomment this line to make nvcc output register usage for each kernel.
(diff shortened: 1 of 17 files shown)


af1990 (approved):
You're right! Now it's easier to understand the meaning of each function.

host-configs/LLNL/lassen-clang@upstream.cmake
```diff
@@ -54,7 +54,7 @@ set(CMAKE_CUDA_FLAGS_RELWITHDEBINFO "-g -lineinfo ${CMAKE_CUDA_FLAGS_RELEASE}" C
 set(CMAKE_CUDA_FLAGS_DEBUG "-g -G -O0 -Xcompiler -O0" CACHE STRING "")
 
 # Uncomment this line to make nvcc output register usage for each kernel.
-# set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --resource-usage" CACHE STRING "" FORCE)
+ set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --resource-usage" CACHE STRING "" FORCE)
```
corbett5, line 57: Maybe comment this out again?

corbett5 (approved):
This looks much clearer to me! Would you mind updating the sphinx documentation in `docs/sphinx/tensorOps.rst`? Specifically the names of the functions and also the paragraph right before examples that begins with
```
LvArray::tensorOps contains a lot of operations and it can be hard to remember the appropriate order to specify the function arguments and size template parameters.
```

src/genericTensorOps.hpp
```diff
@@ -262,19 +262,19 @@ constexpr std::ptrdiff_t SYM_SIZE = ( M * ( M + 1 ) ) / 2;
...
  * @param vector The vector to get the maximum entry of, of length M.
  */
-template< std::ptrdiff_t M, typename VECTOR >
+template< std::ptrdiff_t ISIZE, typename VECTOR >
 LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK inline
 auto maxAbsoluteEntry( VECTOR && vector )
 {
-  static_assert( M > 0, "M must be greater than zero." );
-  internal::checkSizes< M >( vector );
+  static_assert( ISIZE > 0, "M must be greater than zero." );
```
klevzoff, line 273: A lot of these static assert messages keep referencing old parameter names. Would be best to keep them consistent... hopefully it's as simple as `s/M must/ISIZE must/g`, etc.

Also there are still some references in doxygen comment blocks, e.g. see `@param vector` line above. I would actually recommend removing duplication there, i.e. `@tparam ISIZE` line already explains the size of `vector`, no need to type it twice. This one is more annoying as it may no automate well.

klevzoff (approved):

host-configs/LLNL/lassen-clang@upstream.cmake
rrsettgast: I kind of get sick of uncommenting it. Is there an issue with just leaving it on?

corbett5: It spits out a lot of output at build time that can make error messages hard to find. Most of the time I'm building with nvcc I don't want to see that info.

PR #203, opened 2020-10-07
Spack package

corbett5:


docs/sphinx/ArrayOfSets.rst
```diff
@@ -1,5 +1,5 @@
 ###############################################################################
-``Lvarray::ArrayOfSets``
+``LvArray::ArrayOfSets``
 ###############################################################################
 
 The ``LvArray::ArrayOfSets`` is very siilar to the ``LvArray::ArrayOfArrays`` except that the values of the inner arrays are sorted an unique like the ``LvArray::SortedArray``. If you are familiar with both of these classes the functionality of the ``LvArray::ArrayOfSets`` should be pretty straightforward.
```
bmhan12, line 5: "similar"

bmhan12, line 5: "similar"

bmhan12 (approved):
LGTM! Have some minor suggestions.

docs/sphinx/ArrayOfSets.rst
corbett5: Haha thanks!

host-configs/LLNL/quartz-tribol2-clang@10.0.0.cmake
```diff
@@ -0,0 +1,27 @@
+set(CONFIG_NAME "quartz-tribol2-clang@10.0.0" CACHE PATH "")
+
```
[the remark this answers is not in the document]
corbett5: Yeah these are both getting deleted.

docs/sphinx/ArrayOfSets.rst
corbett5: Fixed.

host-configs/LLNL/lassen-xl@2020.08.13.cmake
```diff
@@ -1,56 +0,0 @@
-set(CONFIG_NAME "lassen-xl@2020.08.13" CACHE PATH "")
```
rrsettgast, line 1: what happened to this?

scripts/uberenv/gen_spack_env_script.py
```diff
@@ -0,0 +1,128 @@
...
+#
+# Produced at the Lawrence Livermore National Laboratory
+#
+# LLNL-CODE-716457
+#
+# All rights reserved.
+#
+# This file is part of Ascent.
+#
+# For details, see: http://ascent.readthedocs.io/.
```
rrsettgast, line 12: Does this need to be updated?

scripts/uberenv/gen_spack_env_script.py
```diff
@@ -0,0 +1,128 @@
+###############################################################################
```
rrsettgast, line 1: Did you make any changes to these files?

rrsettgast (approved):
Just a minor concern about copying some of the uberenv scripts. Perhaps they should be maintained separately in a submodule s.t. all the projects can have maintained scripts.

host-configs/LLNL/lassen-xl@2020.08.13.cmake
corbett5: It didn't have an associated set of TPLs, unlike the other LC configs. If you want to build with XL the easiest thing to do is to just use spack to build the dependencies.

corbett5:
> 
> Just a minor concern about copying some of the uberenv scripts. Perhaps they should be maintained separately in a submodule s.t. all the projects can have maintained scripts.

There is actually an uberenv repository (https://github.com/LLNL/uberenv) and this is what I first tried. I didn't wind up modifying any of the existing files but it seems that you need to add files to the uberenv directory. If I remember correctly I could only get it to work when the `project.json`, `packages` and `spack_configs` directories were in the same directory as `uberenv.py`. So instead of maintaining a submodule branch I decided to just do what the other users are doing and embed the files directly. It's entirely possible I was overlooking something and putting those project specific files elsewhere is just a matter of specifying the correct command line arguments. I think at some point the Spack team would like to make uberenv obsolete but I have no idea what that timeline might be.

@white238 @cyrush thoughts?

scripts/uberenv/gen_spack_env_script.py
corbett5: I took these directly from the uberenv repo ((https://github.com/LLNL/uberenv) and didn't make any changes so I'd prefer not to touch them at all.

white238:
> > Just a minor concern about copying some of the uberenv scripts. Perhaps they should be maintained separately in a submodule s.t. all the projects can have maintained scripts.
> 
> There is actually an uberenv repository (https://github.com/LLNL/uberenv) and this is what I first tried. I didn't wind up modifying any of the existing files but it seems that you need to add files to the uberenv directory. If I remember correctly I could only get it to work when the `project.json`, `packages` and `spack_configs` directories were in the same directory as `uberenv.py`. So instead of maintaining a submodule branch I decided to just do what the other users are doing and embed the files directly. It's entirely possible I was overlooking something and putting those project specific files elsewhere is just a matter of specifying the correct command line arguments. I think at some point the Spack team would like to make uberenv obsolete but I have no idea what that timeline might be.
> 
> @white238 @cyrush thoughts?
> 
> Edit:
> I'm going to merge this but there will be a follow up PR once [spack/spack#19221](https://github.com/spack/spack/pull/19221) goes in. If my usage of uberenv could be improved I'll make the changes there.

The current design/grew this way was to have the files embedded like you have done.  This will probably be fixed soonish by Adrien Bernede. We have spoken on some ideas on how to work around requiring these files to be at the same level.

Yes the spack team doesn't like uberenv and the whole point of uberenv was to work around the limitations of Spack while also easing it's use.  The latter will probably never go away due to most users of your code will not want to run/understand multiple spack commands to build your TPLs.

cyrush:
We have talked about projects using uberenv via a submodule a few times.  So far, it seems copying the script is simply the easiest way to use it. 

As you noted, it would require changes to uberenv  - based on where it looks for files etc.

Also, without CI checks, it's pretty easy to accidentally time-travel when jumping around branches.  With build automation stuff (like uberenv) this can lead to lots of head scratching. We face this for BLT in other repos, so have CI checks that make sure we actually intend to change the submodule commit, they look for the sha in another file.

corbett5:
@white238 @cyrush thanks! I'll keep an eye on the uberenv repo and update to a submodule once that's supported.

PR #204, opened 2020-10-09
Corbett/update tpls

corbett5:


rrsettgast (approved):
Why the change of case for raja?

corbett5:
@rrsettgast  
> 
> Why the change of case for raja?

So we can just do `find_package(RAJA)` instead of all the complicated logic that was in `FindRAJA.cmake`.

corbett5:
@af1990 
> 
> I cannot run on Lassen, so I cannot check that the changes in the lassen host-config are working.
> The renaming of RAJA is fine.

The lassen issues don't appear to be due to the TPL update.

corbett5:
Changes already added.

PR #210, opened 2020-10-23
fix doxygen option and path on LC quartz builds

rrsettgast:
https://github.com/GEOSX/GEOSX/pull/1214


CusiniM (approved):
I see, it was the LvArray that was changing the path. You can ignore my comment in the other PR.

corbett5 (approved):

PR #211, opened 2020-10-26
Corbett/rvalue toview

### LINKED ISSUE #168: Add toSlice to SortedArrayView
`SortedArrayView` and by extension `SortedArray` should be able to create an `ArraySlice< T const, 1, 0 >`.

### LINKED ISSUE #197: Disalow creation of views and slices from rvalues.
You should not be able to create an `ArrayView` or an `ArraySlice` from a `Array &&`  and you should not be able to create an `ArraySlice` from an `ArrayView &&`. Similarly for the other classes. This includes the creation of `ArraySlice` from the `operator[]` but not from `CRSMatrixView::getEntries`.

At the same time it would be nice to see if there's an easy way to disable the conversion from `Array` to `ArrayView &`. If this involves making `ArrayView` a protected base class then a separate issue should be made.

corbett5:
Protects against a bunch of rvalue bugs. Not everything is protected though, for example you can still call `data` on an `Array &&`.

Closes #197, closes #168.

Related to https://github.com/GEOSX/GEOSX/pull/1216

### DIFF
--- a/src/ArrayView.hpp
+++ b/src/ArrayView.hpp
@@ -229,17 +229,17 @@ public:
   ///@{
 
   /**
-   * @return Return *this after converting any nested arrays to const views.
+   * @return Return a new ArrayView.
    */
   inline LVARRAY_HOST_DEVICE constexpr
-  ArrayView toView() const
+  ArrayView toView() const &
   { return ArrayView( m_dims, m_strides, m_singleParameterResizeIndex, m_dataBuffer ); }
 
   /**
-   * @return Return *this after converting any nested arrays to const views to const values.
+   * @return Return a new ArrayView where @c T is @c const.
    */
   inline LVARRAY_HOST_DEVICE constexpr
-  ArrayView< T const, NDIM, USD, INDEX_TYPE, BUFFER_TYPE > toViewConst() const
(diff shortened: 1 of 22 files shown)


src/ArrayView.hpp
```diff
@@ -296,28 +295,45 @@ class ArrayView
...
+  /**
+   * @brief Overload for rvalues that raises a compilation error when used.
+   * @return A null ArrayView.
+   * @brief This is valid when this is an @c ArrayView but invalid when this is an @c Array.
+   *   I have been unable to allow the @c ArrayView conversion while not allowing the @c Array.
+   *   Therefore the @c ArrayView conversion is not allowed, instead simply call @c toViewConst.
+   */
+  template< typename _T=T >
+  inline LVARRAY_HOST_DEVICE constexpr
+  operator std::enable_if_t< !std::is_const< _T >::value,
```
corbett5, line 329: This is valid for an `ArrayView` but not an `Array`. I wasn't able to find a solution that allowed this with out errors in either Clang or GCC and warnings in NVCC. I chose to disallow it for `ArrayView` since you can always explicitly call `toViewConst`. This is the source of most of the changes in GEOSX.

rrsettgast (approved):
Looks great

PR #213, opened 2020-11-05
Feature/boomer amg for elasticity

rrsettgast:


src/ArrayOfArraysView.hpp
```diff
@@ -558,6 +579,12 @@ class ArrayOfArraysView
     m_offsets.move( space, touch );
   }
 
+
+  void moveValues( MemorySpace const space, bool touch=true ) const
```
corbett5, line 583: What did you need this for that `move` didn't accomplish?

corbett5 (approved):

src/limits.hpp
```diff
@@ -117,7 +117,8 @@ integerConversion( INPUT input )
   static_assert( std::is_integral< INPUT >::value, "INPUT must be an integral type." );
   static_assert( std::is_integral< OUTPUT >::value, "OUTPUT must be an integral type." );
 
-  return OUTPUT{ input };
+//  return OUTPUT{ input };
+  return static_cast<OUTPUT>(input);
```
corbett5, line 121: The braced initialization broke something?

corbett5:
Oops I just saw this was a WIP, but looks good to me!

src/ArrayOfArraysView.hpp
rrsettgast: Oh....It was to just move the "values" and not any of the other stuff. I think it is unnecessary now. I will probably remove this.

src/limits.hpp
rrsettgast: yes...but I think I fixed it elsewhere. I will try to revert this.

rrsettgast: there was an integer narrowing error.

PR #215, opened 2020-11-17
Bugfix for GCC 9.

### LINKED ISSUE #214: GCC 9.2.1 issue with static_assert
Henri Calandra had issues compiling GEOSX with GCC 9.2.1. The problem were asserts of the form `static_assert( !typeManipulation ::always_true<T>, … )`. The issue appears to be in the implementation of `always_true` whose purpose is to obfuscate the conditional so that the `static_assert` is only triggered if the function is used, not when it's parsed. It appears that the current implementation is too simple.

corbett5:
Closes #214

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,7 @@ Version vxx.yy.zz -- Release date 20yy-mm-dd
   * Fixed indexing bug that produced a compilation error with GCC 10.
   * Fixed a memory leak in `CRSMatrix::assimilate`.
   * Fixed a memory leak in `ArrayOfArraysView::free` with non-trivially destructable values.
+  * Fixed a compilation error with GCC 9.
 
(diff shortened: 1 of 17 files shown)


corbett5:
This could have been a small fix to just small fix to `HAS_MEMBER_FUNCTION_NO_RTYPE` but I realized that just using `foo() const && = delete` is probably the way to go for functions / operators that I want to disallow. You don't get the nice static assert message but it's much cleaner.

I also found a way to allow the UDC from `Array< T > const &` to `ArrayView< T const >` and from `ArrayView< T > &&` to `ArrayView< T const >` while disallowing `Array< T > &&` to `ArrayView< T const >`. Previously all of three were disallowed. The changes to the tests are to prevent the instantiation of `Array< T const >` which now produces a UDC never used warning from NVCC.

rrsettgast (approved):

PR #216, opened 2020-12-09
Add symmetric operations to tensors

rrsettgast:
Symmetric vector outer product
symmetric vector self-outer product

### DIFF
--- a/src/fixedSizeSquareMatrixOps.hpp
+++ b/src/fixedSizeSquareMatrixOps.hpp
@@ -195,6 +195,29 @@ void Rij_eq_AiAj( DST_SYM_MATRIX && LVARRAY_RESTRICT_REF dstMatrix,
                                                vectorA );
 }
 
+/**
+ * @brief Perform the unscaled symmetric outer product of @p vectorA and
+ *   @p vectorB with itself writing the result to @p dstMatrix.
+ * @tparam M The size of both dimensions of @p dstMatrix and the length of @p vectorA.
+ * @tparam DST_MATRIX The type of @p dstMatrix.
+ * @tparam VECTOR_A The type of @p vectorA.
+ * @tparam VECTOR_B The type of @p vectorB.
+ * @param dstMatrix The matrix the result is written to, of size M x N.
+ * @param vectorA The first vector in the outer product, of length M.
+ * @param vectorB The second vector in the outer product, of length M.
+ * @details Performs the operations @code dstMatrix[ i ][ j ] = vectorA[ i ] * vectorB[ j ] + vectorA[ j ] * vectorB[ i ] @endcode
+ */
+template< std::ptrdiff_t M, typename DST_SYM_MATRIX, typename VECTOR_A, typename VECTOR_B >
+LVARRAY_HOST_DEVICE CONSTEXPR_WITHOUT_BOUNDS_CHECK inline
+void Rij_eq_AiBj_plus_AjBi( DST_SYM_MATRIX && LVARRAY_RESTRICT_REF dstMatrix,
+                            VECTOR_A const & LVARRAY_RESTRICT_REF vectorA,
+                            VECTOR_B const & LVARRAY_RESTRICT_REF vectorB )
+{
+  internal::SquareMatrixOps< M >::Rij_eq_AiBj_plus_AjBi( std::forward< DST_SYM_MATRIX >( dstMatrix ),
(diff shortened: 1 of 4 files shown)


unitTests/testTensorOpsFixedSize.cpp
```diff
@@ -351,6 +351,113 @@ class FixedSizeSquareMatrixTest : public ::testing::Test
...
+    std::ptrdiff_t const matrixSeed = m_seedSymMatrixA;
+
+    forall< POLICY >( 1, [symResult, symMatrixA_IJ, symMatrixA_JI, vectorA_IJ, vectorA_JI, vectorA_local, matrixSeed]
+                      LVARRAY_HOST_DEVICE ( int )
+        {
+          #define _TEST( symMatrix, vectorN ) \
+            fill( symMatrix, matrixSeed ); \
+            tensorOps::symRij_eq_AiAj< N >( symMatrix, vectorN ); \
+            CHECK_EQUALITY_1D( SYM_SIZE, symMatrix, symResult )
+
```
CusiniM, line 384: @corbett5 since that matrix gets overwritten I wonder if it is necessary here to use all the permutations of `symMatrix` and also if that `fill` is actually necessary.

unitTests/testTensorOpsTwoSizes.hpp
```diff
@@ -669,6 +669,65 @@ class TwoSizesTest : public ::testing::Test
...
+
+    std::ptrdiff_t const matrixSeed = m_matrixASeed;
+
+    forall< POLICY >( 1, [matrixA_IJK, matrixA_IKJ, matrixA_KJI, matrixB_IJK_view, matrixB_IKJ_view, matrixB_KJI_view, matrixB_local, matrixSeed ]
+                      LVARRAY_HOST_DEVICE ( int )
+        {
+          #define _TEST( dstMatrix, srcMatrix ) \
+            fill( dstMatrix, matrixSeed ); \
+            tensorOps::scaledAdd< N, M >( dstMatrix, srcMatrix, scale ); \
+            CHECK_EQUALITY_2D( N, M, dstMatrix, result ); \
```
CusiniM, line 692: @corbett5 the vector version of this function uses a tolerance `CHECK_NEAR_1D`. Is this going to be ok? `CHECK_NEAR_2D` does not seem to exist.

unitTests/testTensorOpsFixedSize.cpp
corbett5: The call to `fill` is necessary because it `symMatrix` is re-used between calculations each of which should produce the same result. If there was no call to fill, the first calculation might work but then subsequent ones could do nothing and the tests would pass.

As for the permutations they're mostly there for compile time checks.

unitTests/testTensorOpsTwoSizes.hpp
corbett5: Ideally they would all use `CHECK_EQUALITY_*`, but FMA gives different answers with some compilers and devices. Can you try it out on Lassen with both clang and GCC in release and debug? If that works then it's fine with me, otherwise adding a `CHECK_NEAR_2D` should be pretty easy.

corbett5 (approved):
Looks good, thanks!

unitTests/testTensorOpsTwoSizes.hpp
```diff
@@ -669,6 +669,119 @@ class TwoSizesTest : public ::testing::Test
...
+          _TEST_PERMS( matrixA_local, matrixB_IJK_view[ 0 ], matrixB_IKJ_view[ 0 ], matrixB_KJI_view[ 0 ], matrixB_local );
+          _TEST_PERMS( matrixA_local, matrixB_IJK_view[ 0 ], matrixB_IKJ_view[ 0 ], matrixB_KJI_view[ 0 ], matrixB_local );
+          _TEST_PERMS( matrixA_local, matrixB_IJK_view[ 0 ], matrixB_IKJ_view[ 0 ], matrixB_KJI_view[ 0 ], matrixB_local );
+
+            #undef _TEST_PERMS
+            #undef _TEST
+        } );
+  }
+
+  void testAkiAkj()
```
francoishamon, line 731: @corbett5 I am done. My changes are in this function if you want to double check. Thanks

corbett5: Looks great, thanks!

PR #220, opened 2021-01-21
Using ChaiBuffer to test on GPUs

TotoGaz:
Testing `exampleArrayOfArrays` on Pangea3 without magical memory exchange between GPU and CPU emphasizes that the wrong buffer type is used in the test.

Related to https://github.com/GEOSX/GEOSX/issues/1278

### DIFF
--- a/examples/exampleArrayOfArrays.cpp
+++ b/examples/exampleArrayOfArrays.cpp
@@ -279,14 +279,14 @@ TEST( ArrayOfArrays, resizeFromCapacities )
 // Sphinx start after ChaiBuffer
 CUDA_TEST( ArrayOfArrays, ChaiBuffer )
 {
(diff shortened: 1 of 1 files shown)


TotoGaz:
I'm not sure about the merging process with the `submodule` stuff in GEOSX.
Do I have to make a PR in GEOSX too?

corbett5 (approved):

corbett5:
Yes. In this case I would just merge this and then create a GEOSX PR with the update. In cases where the changes here interact with GEOSX you should create a PR here and one in GEOSX with the updated submodule and any associated code changes. Then once both PRs are approved, merge the submodule, update the submodule in the GEOSX branch and then merge that PR.

As a tip you merge the submodule PR by doing `git push my-branch:develop`  (or something like that and assuming your branch is up-to-date with `develop`) from the command line which will mean that you won't have to update the GEOSX PR since the commit hash won't change, although you may have to re-run the submodule check test.

PR #224, opened 2021-02-12
Added exception macros.

corbett5:


src/Macros.hpp
```diff
@@ -138,13 +138,28 @@
...
+    { \
+      std::ostringstream __oss; \
+      __oss << "\n"; \
+      __oss << "***** LOCATION: " LOCATION "\n"; \
+      __oss << "***** Controlling expression (should be false): " STRINGIZE( EXP ) "\n"; \
+      __oss << MSG << "\n"; \
+      __oss << LvArray::system::stackTrace( true ); \
+      throw TYPE( __oss.str() ); \
+    } \
+  } while( false )
```
TotoGaz, line 154: Why do you actually need to embed this in some `do { ... } while( false )`?
It's the same pattern as `LVARRAY_ERROR_IF` but I do not understand anyway.
For scope protection a pair of curly brackets would be enough, wouldn't they? (I'm not even sure it's needed since the `if` defines its own scope)

corbett5: It's so that you are forced to put a semicolon after the macro. I just tried it with clang 10.0.0 and gcc 8.1.0 and neither of them care (at least with our flags) if you have put a semi-colon after an `if` block, but I think other versions might. Either way I think it's nice to force a semicolon.

corbett5:
LGTM

rrsettgast (approved):

PR #226, opened 2021-03-04
Fix warnings in Doxygen comments

klevzoff:
Related to https://github.com/GEOSX/GEOSX/pull/1345


corbett5 (approved):
My bad for not documenting those, thanks!

PR #230, opened 2021-04-03
Feature/han12/combine static

rrsettgast:


CMakeLists.txt
```diff
@@ -35,8 +35,12 @@ if( NOT is_submodule )
     include( cmake/blt/SetupBLT.cmake )
     include( cmake/CMakeBasics.cmake )
     include( cmake/SetupTPL.cmake )
+    set( LvArrayBuildAsObj FALSE)
+    
 else()
     include( cmake/CMakeBasics.cmake )
```
corbett5, line 41: Could we make this an `option( LVARRAY_BUILD_OBJECT_LIBRARIES OFF )` and then have GEOSX just set it before including?

corbett5 (approved):

CMakeLists.txt
rrsettgast: of course

Issue #172, opened 2020-07-07
Use the camp platforms instead of MemorySpace


corbett5:
https://github.com/LLNL/camp/blob/master/include/camp/resource/platform.hpp
https://github.com/LLNL/RAJA/blob/develop/include/RAJA/internal/get_platform.hpp
https://github.com/LLNL/RAJA/blob/develop/include/RAJA/util/camp_aliases.hpp

corbett5:
Move `bufferManipulation::HasMemberFunction_move` into `typeManipulation.hpp`.

PR #231, opened 2021-04-23
0.2.0 release

corbett5:
Creates a new release as well as adds a bunch of new features

- You can specify the allocators a specific `ChaiBuffer` should use, and construct an `Array` with that buffer.
- You can allocate an `Array` directly on device. This only works when the `Array` was previously empty, but this should be the most common use case.
- Adds half precision math functions.
- Added a conversion from `ArrayView< T, N, NDIM >` to `ArrayView< U, N, NDIM >`, the use case I had was to convert between arrays of `__half` and `__half2`.
- Improved the error logging when you have an `LVARRAY_ERROR` (or `ASSERT`) on device.
- Added functions to `memcpy` an `ArraySlice`, it uses Umpire so it works between memory spaces.
- Added a `zero` method to `ArrayView` and `CRSMatrixView` that will use `memset` instead of launching a kernel to zero it out.
- `arrayManipulation::resize` (which is used by pretty much every class) now uses `memset` when it will be zero initializing memory. This should be a lot faster for large arrays, it's possible it will be slower for really small things but it would be easy enough to add in a different path for small cases.
- Replace the `MemorySpace` enum with `using MemorySpace = camp::resource::Platform`.


corbett5:
Closes #172.

corbett5:
Both Spheral and Tribol would like a new release, plus the original release didn't have the Python stuff.

src/Array.hpp
```diff
@@ -439,7 +473,7 @@ class Array : public ArrayView< T,
    *        capacity() >= newCapacity.
    */
   void reserve( INDEX_TYPE const newCapacity )
-  { bufferManipulation::reserve( this->m_dataBuffer, this->size(), newCapacity ); }
+  { bufferManipulation::reserve( this->m_dataBuffer, this->size(), MemorySpace::host, newCapacity ); }
```
rrsettgast, line 476: shouldn't the space be propagated to the outer interface?

rrsettgast (approved):

rrsettgast:
@corbett5 Don't create the release until we setup the auto DOI generation.

src/ArrayView.hpp
```diff
@@ -545,18 +602,39 @@ class ArrayView
...
       } );
   }
 
+  /**
+   * @brief Use memset to set all the values in the array to 0.
+   * @details This is preferred over setValues< POLICY >( 0 ) for numeric types since it is much faster in most cases.
+   *   If the buffer is allocated using Umpire then the Umpire ResouceManager is used, otherwise std::memset is used.
+   * @note The memset occurs in the last space the array was used in and the view is moved and touched in that space.
+   */
+  inline void zero() const
```
rrsettgast, line 625: do we want to allow setting the memory space?

scripts/uberenv/packages/lvarray/package.py
```diff
@@ -70,7 +74,7 @@ class Lvarray(CMakePackage, CudaPackage):
 
     depends_on('python +shared +pic', when='+pylvarray')
     depends_on('py-numpy@1.19: +blas +lapack +force-parallel-build', when='+pylvarray')
-    depends_on('py-scipy@1.5.2: +force-parallel-build', when='+pylvarray')
+    # depends_on('py-scipy@1.5.2: +force-parallel-build', when='+pylvarray')
```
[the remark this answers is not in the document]
corbett5: No, thanks!

src/Array.hpp
corbett5: I don't think so. At the moment the only way to allocate on device (without moving from host) is call `resizeWithoutInitializationOrDestruction` on an empty Array. If the Array already has an allocation (on host or device) it won't work. This is probably the majority use case, and if you want to resize the device only Array you can create a new device only Array and copy the values over. Doing a more general resize/reallocation on device without launching a kernel is tricky and not possible in all cases such as any time you have a default value. I'd certainly like to expand this capability and as soon as there's a use case it should be easy to add.

Doing the single parameter resize where the resize dimension is not the largest stride is serial (basically inserting new values in a bunch of places) and doing so on device would involve either creating a new device allocation each time or moving it back to host and doing the work there.

src/ArrayView.hpp
corbett5: Definitely since it can avoid a move. I think the best way to do this is have `registerTouch` allocate in the given space if it didn't already exist and then touch it there. That way all `memcpy` has to do is replace `move( space, true )` with `registerTouch( space )`. I don't really want to add that to this PR though so I'll create an issue.

scripts/uberenv/packages/lvarray/package.py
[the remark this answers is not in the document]
rrsettgast: I don't understand. Was the exclusion of scipy intentional? If this is the python that must be used when interfacing with LvArray, then should we be including a large collection of packages like scipy, matplotlib...etc.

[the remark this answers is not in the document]
corbett5: No, I was testing out the NumPy build on Lassen which I thought I had working with ESSL but I guess not. SciPy takes a bit to build so I commented it out. We should only list packages here that need to be built from source, everything else should and can be pip installed.

PR #236, opened 2021-06-21
Make calling begin/end on non-contiguous slices a compile-time error

klevzoff:
This PR turns calling `begin`/`end` on a slice that lost its USD into a compile-time error (previously a guaranteed runtime error).

Related to https://github.com/GEOSX/GEOSX/pull/1463

### DIFF
--- a/src/ArraySlice.hpp
+++ b/src/ArraySlice.hpp
@@ -141,7 +141,6 @@ public:
   /**
    * @return Return a new immutable slice.
    */
-  template< typename U=T >
   LVARRAY_HOST_DEVICE inline constexpr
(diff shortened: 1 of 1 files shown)


corbett5 (approved):
I know in another PR you mentioned casting from a slice with known `USD` to a slice where `USD = -1` in order to reduce the combinatorial explosion. This would break that, or at least limit the use of these methods. That's fine with me since as you mentioned it isn't supported anyways.

Anyways I think it might be nice to add specializations for `USD < 0` that have a `static_assert(USD < 0, "Not available on non-contiguous slices")` just to make it a bit easier for users.

PR #239, opened 2021-07-29
Kahan 2x2 determinant computation reduces numerical imprecision.

TotoGaz:
Related to issue https://github.com/GEOSX/GEOSX/issues/1440: some tests `testTensorOpsInverseTwoArgs` and `testTensorOpsInverseOneArg` are failing due to large numerical imprecision.

Using the Kahan method to compute 2x2 determinants brings more precision.
3x3 case still untouched.

Warning, I sometimes get compilation warnings like
```
../coreComponents/LvArray/unitTests/../src/fixedSizeSquareMatrixOpsImpl.hpp(100): warning: calling a constexpr __host__ function("fma") from a __host__ __device__ function("determinant") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
```
which I find surprising since this `fma` function is supposed to be `device` friendly.
`fma` also [seems pretty standard](https://www.cplusplus.com/reference/cmath/fma/) so one should not need any fallback.
So I need to double-check.

Any opinion?

### DIFF
--- a/src/fixedSizeSquareMatrixOpsImpl.hpp
+++ b/src/fixedSizeSquareMatrixOpsImpl.hpp
@@ -89,17 +89,9 @@ struct SquareMatrixOps< 2 >
   {
     checkSizes< 2, 2 >( matrix );
 
-    // Aliases for matrix [[a, b],[c, d]] for improved readability
-    auto const & a = matrix[0][0];
-    auto const & b = matrix[0][1];
-    auto const & c = matrix[1][0];
-    auto const & d = matrix[1][1];
-
     // Using the more precise Kahan method to compute the 2x2 determinant.
-    auto const w = b * c;
-    auto const e = fma( -b, c, w );
-    auto const f = fma( a, d, -w );
-    return f + e;
+    auto const tmp = matrix[0][1] * matrix[1][0];
+    return math::fma( -matrix[0][1], matrix[1][0], tmp ) + math::fma( matrix[0][0], matrix[1][1], -tmp );
   }
 
   /**
--- a/src/fixedSizeSquareMatrixOpsImpl.hpp
+++ b/src/fixedSizeSquareMatrixOpsImpl.hpp
@@ -88,7 +88,18 @@ struct SquareMatrixOps< 2 >
   static auto determinant( MATRIX const & matrix )
   {
     checkSizes< 2, 2 >( matrix );
-    return matrix[ 0 ][ 0 ] * matrix[ 1 ][ 1 ] - matrix[ 0 ][ 1 ] * matrix[ 1 ][ 0 ];
+
+    // Aliases for matrix [[a, b],[c, d]] for improved readability
+    auto const & a = matrix[0][0];
+    auto const & b = matrix[0][1];
+    auto const & c = matrix[1][0];
+    auto const & d = matrix[1][1];
+
+    // Using the more precise Kahan method to compute the 2x2 determinant.
+    auto const w = b * c;
+    auto const e = fma( -b, c, w );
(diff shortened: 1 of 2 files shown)


src/fixedSizeSquareMatrixOpsImpl.hpp
```diff
@@ -88,7 +88,18 @@ struct SquareMatrixOps< 2 >
   static auto determinant( MATRIX const & matrix )
   {
     checkSizes< 2, 2 >( matrix );
-    return matrix[ 0 ][ 0 ] * matrix[ 1 ][ 1 ] - matrix[ 0 ][ 1 ] * matrix[ 1 ][ 0 ];
+
+    // Aliases for matrix [[a, b],[c, d]] for improved readability
+    auto const & a = matrix[0][0];
```
corbett5, line 93: I think `matrix[ 0 ][ 0 ]` is more readable.

src/fixedSizeSquareMatrixOpsImpl.hpp
```diff
@@ -88,7 +88,18 @@ struct SquareMatrixOps< 2 >
...
+    auto const & b = matrix[0][1];
+    auto const & c = matrix[1][0];
+    auto const & d = matrix[1][1];
+
+    // Using the more precise Kahan method to compute the 2x2 determinant.
+    auto const w = b * c;
+    auto const e = fma( -b, c, w );
```
corbett5, line 100: Is the `fma` required to obtain adequate precision? I'd prefer to do `auto const e = -b * c + w` and let the compiler do it's thing.

corbett5 (approved):
If we want to use `fma` we should put it in `LvArray::math` and wrap it such that it uses `std::fma` on host and CUDA's `fma` on device.

Also this modification makes calculating the determinant significantly more costly. I'm not sure you can answer this question but maybe the kind of matrices that pop up in the unit tests aren't representative of the normal use case.

src/fixedSizeSquareMatrixOpsImpl.hpp
TotoGaz: OK, I will change this.

src/fixedSizeSquareMatrixOpsImpl.hpp
TotoGaz: Sure, I'll check too.

TotoGaz:
> Also this modification makes calculating the determinant significantly more costly. I'm not sure you can answer this question but maybe the kind of matrices that pop up in the unit tests aren't representative of the normal use case.

I could spot 3x3 dets in the code (FEM stuff), but only 2x2 in the unit tests.
I will double check.

src/fixedSizeSquareMatrixOpsImpl.hpp
rrsettgast: https://godbolt.org/z/ao9GMhPsq

src/fixedSizeSquareMatrixOpsImpl.hpp
rrsettgast: > Also this modification makes calculating the determinant significantly more costly. I'm not sure you can answer this question but maybe the kind of matrices that pop up in the unit tests aren't representative of the normal use case.

It is 3 flop vs 2 flop. Look at the second variant I listed in https://godbolt.org/z/4bGdjYbd5

I don't think this is a big deal...for 3d it may be a bit more costly...but not much. I don't think we will often run into this sort of round off problem, but then again, if it doesn't cost much to "fix", then I kind of like this approach.

PR #242, opened 2021-08-29
Corbett/fix leaks

corbett5:


CTestConfig.cmake
```diff
@@ -0,0 +1,3 @@
+set(MEMORYCHECK_COMMAND_OPTIONS "--tool=memcheck --partial-loads-ok=yes --error-limit=no --leak-check=full --show-reachable=yes --max-stackframe=16777216 --num-callers=20 --child-silent-after-fork=yes --track-origins=yes --error-exitcode=1 --errors-for-leak-kinds=all")
```
corbett5, line 1: @rrsettgast if you add this file at the top level you can configure how CTest will run valgrind. Then  you can do `ctest  -T memcheck` to run all the tests through valgrind. Or you can combine it with `-V -R` to run a specific test. It does require having a very robust suppression file, but you can fix that by removing `--errors-for-leak-kinds=all`

rrsettgast (approved):

PR #244, opened 2021-09-18
Feature/settgast/wellbore mesh2

rrsettgast:
1. Update TPL locations on LC
2. fix bug with memory space in reallocation
3. modifications/workarounds for summit build (cuda11)


src/typeManipulation.hpp
```diff
@@ -430,6 +428,15 @@ template< camp::idx_t... INDICES >
...
 { return sizeof...( INDICES ); }
 
+} // namespace internal
+
+/**
+ * @tparam T The type for which the number of indices is requested.
+ * @return the number of indices for type T
+ */
+template< typename T >
+static constexpr camp::idx_t getDimension = internal::getDimension( T {} );
```
klevzoff, line 438: As this is not a function, `@return` is probably not necessary; but if doxygen is happy, no need to change. Also I don't think `static` is doing anything here, but it also doesn't hurt.

klevzoff (approved):

src/typeManipulation.hpp
rrsettgast: I think there might be a linkage issue without static?

https://stackoverflow.com/questions/50488831/use-of-constexpr-in-header-file

or I think that is only for `const` and `constexpr` takes care of it.

PR #248, opened 2022-02-23
Changes to build on crusher with hip+rocm dependencies

wrtobin:
No in-code changes other than accomodating TPL version differences


corbett5:
Hey Bill, any chance you could add HIP support to the LvArray spack build? I would do it but I don't have access to any HIP machines. I bet with my (or @bmhan12's) help it would only take a few hours.

src/Macros.hpp
```diff
@@ -120,20 +148,20 @@
...
-      __oss << "***** ERROR\n"; \
-      __oss << "***** LOCATION: " LOCATION "\n"; \
-      __oss << "***** Controlling expression (should be false): " STRINGIZE( EXP ) "\n"; \
-      __oss << MSG << "\n"; \
-      __oss << LvArray::system::stackTrace( true ); \
-      std::cout << __oss.str() << std::endl; \
-      LvArray::system::callErrorHandler(); \
-    } \
-  } while( false )
+  // do \
```
corbett5: I assume you commented this out for some strange HIP compilation reason. Just a reminder (to me as well) that this definitely needs to be fixed.

rrsettgast (approved):

PR #256, opened 2022-03-30
Enhancements for ArrayOf[Arrays|Sets] construction/assimilation

klevzoff:
Related to: https://github.com/GEOSX/GEOSX/pull/1847
  * Add execution policy template parameter to `ArrayOfSets::assimilate()` (can be host or device policy).
  * Add new method `ArrayOfArrays::resizeFromOffsets()`. This is useful when a new `ArrayOfArrays` needs to "inherit" capacity allocations from another existing one (can be extracted via `ArrayOfArraysView::getOffsets()`), but otherwise be resized to empty state.

### DIFF
--- a/src/ArrayOfArraysView.hpp
+++ b/src/ArrayOfArraysView.hpp
@@ -15,6 +15,7 @@
 // Source includes
 #include "bufferManipulation.hpp"
 #include "arrayManipulation.hpp"
+#include "sortedArrayManipulation.hpp"
 #include "ArraySlice.hpp"
 #include "typeManipulation.hpp"
 #include "math.hpp"
@@ -720,15 +721,48 @@ protected:
     // RAJA::inclusive_scan fails on empty input range
     if( numSubArrays > 0 )
     {
-      // const_cast needed until for RAJA bug.
-      RAJA::inclusive_scan< POLICY >( const_cast< INDEX_TYPE * >( capacities ),
-                                      const_cast< INDEX_TYPE * >( capacities + numSubArrays ),
+      RAJA::inclusive_scan< POLICY >( capacities,
+                                      capacities + numSubArrays,
                                       m_offsets.data() + 1 );
     }
 
     m_numArrays = numSubArrays;
     INDEX_TYPE const maxOffset = m_offsets[ m_numArrays ];
     typeManipulation::forEachArg( [ maxOffset] ( auto & buffer )
+    {
+      bufferManipulation::reserve( buffer, 0, MemorySpace::host, maxOffset );
+    }, m_values, buffers ... );
+  }
+
+  /**
+   * @brief Clears the array and creates a new array with the given number of sub-arrays.
+   * @param numSubArrays The new number of arrays.
+   * @param offsets A pointer to an array of length @p numSubArrays+1 containing the offset
+   *   of each new sub array. Offsets are precomputed by the caller.
+   * @param buffers A variadic pack of buffers to treat similarly to m_values.
+   */
+  template< typename ... BUFFERS >
+  void resizeFromOffsets( INDEX_TYPE const numSubArrays,
+                          INDEX_TYPE const * const offsets,
+                          BUFFERS & ... buffers )
+  {
+    LVARRAY_ASSERT( arrayManipulation::isPositive( numSubArrays ) );
+    LVARRAY_ASSERT_EQ( offsets[0], 0 );
+    LVARRAY_ASSERT( sortedArrayManipulation::isSorted( offsets, offsets + numSubArrays + 1 ) );
+
+    destroyValues( 0, m_numArrays, buffers ... );
+
+    bufferManipulation::reserve( m_sizes, m_numArrays, MemorySpace::host, numSubArrays );
+    std::fill_n( m_sizes.data(), numSubArrays, 0 );
+
+    INDEX_TYPE const offsetsSize = ( m_numArrays == 0 ) ? 0 : m_numArrays + 1;
+    bufferManipulation::reserve( m_offsets, offsetsSize, MemorySpace::host, numSubArrays + 1 );
+
+    arrayManipulation::uninitializedCopy( offsets, offsets + numSubArrays + 1, m_offsets.data() );
+
+    m_numArrays = numSubArrays;
+    INDEX_TYPE const maxOffset = m_offsets[ m_numArrays ];
+    typeManipulation::forEachArg( [ maxOffset ] ( auto & buffer )
     {
       bufferManipulation::reserve( buffer, 0, MemorySpace::host, maxOffset );
     }, m_values, buffers ... );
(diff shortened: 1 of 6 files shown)


src/ArrayOfArraysView.hpp
```diff
@@ -720,9 +721,8 @@ class ArrayOfArraysView
     // RAJA::inclusive_scan fails on empty input range
     if( numSubArrays > 0 )
     {
-      // const_cast needed until for RAJA bug.
-      RAJA::inclusive_scan< POLICY >( const_cast< INDEX_TYPE * >( capacities ),
-                                      const_cast< INDEX_TYPE * >( capacities + numSubArrays ),
```
klevzoff, line 725: Casts seem no longer needed with current version of RAJA we're using

src/ArrayOfArraysView.hpp
```diff
@@ -734,6 +734,40 @@ class ArrayOfArraysView
...
+   * @brief Clears the array and creates a new array with the given number of sub-arrays.
+   * @param numSubArrays The new number of arrays.
+   * @param offsets A pointer to an array of length @p numSubArrays+1 containing the offset
+   *   of each new sub array. Offsets are precomputed by the caller.
+   * @param buffers A variadic pack of buffers to treat similarly to m_values.
+   */
+  template< typename ... BUFFERS >
+  void resizeFromOffsets( INDEX_TYPE const numSubArrays,
+                          INDEX_TYPE const * const offsets,
+                          BUFFERS & ... buffers )
```
klevzoff: I still need to add a unit test for it. But I'd like to get the idea approved first. Usage is as follows:
```c++
newAofA.resizeFromOffsets( otherAofA.size(), otherAofA.toView().getOffsets() );
```

We could also implement `duplicateSparsity` instead, taking a whole AofA (view) as input. Let me know your preferences. It's not clear to me though, whether that method would need to also copy sizes and default construct values. I wouldn't want that to be the case, the new object is meant to be filled via `emplaceBack`s, not overwriting.

src/ArrayOfSets.hpp
```diff
@@ -133,38 +134,43 @@ class ArrayOfSets : protected ArrayOfSetsView< T, INDEX_TYPE, BUFFER_TYPE >
     ParentClass::assimilate( reinterpret_cast< ParentClass && >( src ) );
 
     INDEX_TYPE const numSets = size();
+    ArrayOfArraysView< T, INDEX_TYPE const, true, BUFFER_TYPE > const view =
+      ArrayOfArraysView< T, INDEX_TYPE const, true, BUFFER_TYPE >( numSets,
+                                                                   this->m_offsets,
+                                                                   this->m_sizes,
+                                                                   this->m_values );
+    BUFFER_TYPE< INDEX_TYPE > const sizes = this->m_sizes;
```
klevzoff, line 142: There seems to be no way to manually "resize" a single sub-array through `view` after value deduplication. So I just capture `sizes` separately and overwrite them, seems to work well.

unitTests/testArrayOfSets.cpp
```diff
@@ -890,6 +828,47 @@ class ArrayOfSetsViewTest : public ArrayOfSetsTest< typename ARRAY_OF_SETS_POLIC
     COMPARE_TO_REFERENCE
   }
 
+  // This is not a View test, but it needs a POLICY type argument, so it's placed here for convenience
+  void assimilate( IndexType const maxValue, IndexType const maxInserts, sortedArrayManipulation::Description const desc )
```
klevzoff, line 832: Lmk if you'd like a separate test sub-class to contain these assimilation tests (I moved them from `ArrayOfSetsTest` to `ArrayOfSetsViewTest` for reason stated in the code comment).

src/ArrayOfArraysView.hpp
corbett5: Yeah now that I think about it (in the context of a `CRSMatrix` at least) `duplicateSparsity` would have a different meaning. You'd basically copy all the data wholesale, except set the new values to zero. `resizeFromOffsets` seems like the correct name, although I'm not a huge fan of passing the pointer and size around, but it's not a big deal.

src/ArrayOfArraysView.hpp
```diff
@@ -734,6 +734,40 @@ class ArrayOfArraysView
...
+  {
+    LVARRAY_ASSERT( arrayManipulation::isPositive( numSubArrays ) );
+    LVARRAY_ASSERT_EQ( offsets[0], 0 );
+    LVARRAY_ASSERT( sortedArrayManipulation::isSorted( offsets, offsets + numSubArrays + 1 ) );
+
+    destroyValues( 0, m_numArrays, buffers ... );
+
+    bufferManipulation::reserve( m_sizes, m_numArrays, MemorySpace::host, numSubArrays );
+    std::fill_n( m_sizes.data(), numSubArrays, 0 );
```
corbett5: I know this is just copied from above but maybe this should use `umpireInterface::memset`. I don't think it's important at the moment but if eventually we want to be able to do this without allocating anything on host it'll be necessary.

src/ArrayOfArraysView.hpp
```diff
@@ -720,9 +721,8 @@ class ArrayOfArraysView
     // RAJA::inclusive_scan fails on empty input range
     if( numSubArrays > 0 )
     {
-      // const_cast needed until for RAJA bug.
-      RAJA::inclusive_scan< POLICY >( const_cast< INDEX_TYPE * >( capacities ),
-                                      const_cast< INDEX_TYPE * >( capacities + numSubArrays ),
+      RAJA::inclusive_scan< POLICY >( capacities,
```
corbett5, line 728: This function should probably call `resizeFromOffsets` once the scan has been performed I think.

src/ArrayOfSets.hpp
corbett5: This looks good to me.

unitTests/testArrayOfSets.cpp
corbett5: Nope that sounds just fine. I've done similar things elsewhere.

corbett5 (approved):

src/ArrayOfArraysView.hpp
klevzoff: Ok, done

src/ArrayOfArraysView.hpp
klevzoff: So I couldn't just call it, because the result of the scan is written straight into `m_offsets`, but its old values are still needed in `resizeFromOffsets`, plus it would then perform a self-copy of `m_offsets`. I ended up pulling all the shared code into a private function `resizeFromOffsetsImpl`, which both `resizeFromOffsets` and `resizeFromCapacities` delegate to, passing a callback that is invoked when new offsets must be populated.

PR #261, opened 2022-06-07
Move GPU data to host when converting to numpy array

XL64:


src/python/PyArray.hpp
```diff
@@ -208,7 +208,10 @@ class PyArrayWrapper final : public PyArrayWrapperBase
 
   /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
   virtual PyObject * toNumPy() const final override
-  { return toNumPyImpl(); };
+  {
+    m_array.move(MemorySpace::host, (m_accessLevel == static_cast< int >(LvArray::python::PyModify::READ_ONLY))?false:true);
```
corbett5: I'd rewrite the boolean condition as `m_accessLevel !=  static_cast< int >(PyModify::READ_ONLY)`, getting rid of the ternary and namespace.

corbett5 (approved):

src/python/PyArray.hpp
XL64: done, thanks.

XL64: That's strange, I changed it it is ok on the branch, but not in Files Changed (but if you do show File it will be correct)...

PR #262, opened 2022-06-07
Update/hypre chai raja

rrsettgast:
update hypre/chai/umpire/raja
add clang13 build on LC


corbett5 (approved):
These changes still work with the older versions?

rrsettgast:
I don't think so.

corbett5:
Could you `#ifdef` on the version then? I don't think we want to make it a hard requirement (yet at least).

PR #264, opened 2022-08-25
Spack Update

bmhan12:
Part of PR GEOSX/GEOSX#2028, same type of error as in issue GEOSX/GEOSX#1960 (not sure if this is the fix, error was seen on blueos, original issue is on Pangea3).

Fixes a failing test case with CUDA 11 (issue occurs for negative 0.00)


bmhan12:
Fixed by #265

PR #265, opened 2022-08-26
perform difference modulo 2PI on atan2 testings

XL64:
On Pangea3 testing of math.h were failing because 0 != 4pi.


corbett5 (approved):
LGTM

PR #266, opened 2022-09-01
added equal assignment operator Array<T>::operator=(ArrayView<T const> &)

rrsettgast:
Title says it all.

### DIFF
--- a/src/ArrayView.hpp
+++ b/src/ArrayView.hpp
@@ -85,6 +85,10 @@ public:
   /// The integer type used for indexing.
   using IndexType = INDEX_TYPE;
 
+  /// The type when the data type is const.
+  using ViewTypeConst = ArrayView< std::remove_const_t< T > const, NDIM, USD, INDEX_TYPE, BUFFER_TYPE >;
+
+
   /// The type when all inner array classes are converted to const views.
   using NestedViewType = ArrayView< std::remove_reference_t< typeManipulation::NestedViewType< T > >,
                                     NDIM, USD, INDEX_TYPE, BUFFER_TYPE >;
@@ -274,12 +278,12 @@ public:
    * @return Return a new ArrayView where @c T is @c const.
    */
   inline LVARRAY_HOST_DEVICE constexpr
-  ArrayView< T const, NDIM, USD, INDEX_TYPE, BUFFER_TYPE > toViewConst() const &
+  ViewTypeConst toViewConst() const &
   {
-    return ArrayView< T const, NDIM, USD, INDEX_TYPE, BUFFER_TYPE >( m_dims,
-                                                                     m_strides,
-                                                                     m_singleParameterResizeIndex,
-                                                                     m_dataBuffer );
+    return ViewTypeConst( m_dims,
+                          m_strides,
+                          m_singleParameterResizeIndex,
+                          m_dataBuffer );
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/ArrayView.hpp
```diff
@@ -85,6 +85,10 @@ class ArrayView
   /// The integer type used for indexing.
   using IndexType = INDEX_TYPE;
 
+  /// The type when the data type is const.
+  using ViewTypeConst = ArrayView< std::remove_const_t< T > const, NDIM, USD, INDEX_TYPE, BUFFER_TYPE >;
```
corbett5, line 89: Just curious but is the `std::remove_const` necessary? I thought you don't have problems with duplicate const.

corbett5 (approved):
Looks good to me. I'd either add a test or at least put in a TODO or issue that we should add a test.

klevzoff (approved):

src/ArrayView.hpp
rrsettgast: I had put in `ArrayVIew<T>::operator=( ArrayView<T> const&)`, `ArrayVIew<T>::operator=( ViewTypeConst const &)`, but then thought better of it. I felt it was a little unclear what the intent of the operator is, and whether it would be a source of bugs. I did put in the `std::remove_const` because I was getting some errors in the `using`, but now I am not sure I was this is what fixed the problem I was having. Also:

https://stackoverflow.com/questions/5781222/duplicate-const-qualifier-allowed-in-c-but-not-in-c

but when I remove the `std::remove_const` it compiles fine now. I could remove it, but I feel it is better to be explicit about it...not that this is valid reasoning.

rrsettgast: ...but then again, if I do something like this:
```
  array1d<double> a;
  arrayView1d<double const> aView = a;
  arrayView1d<double const>::ViewTypeConst anotherView = a;

  std::cout<<LvArray::system::demangle(typeid(anotherView).name())<<std::endl;
  ```
it outputs:
```
LvArray::ArrayView<double const, 1, 0, int, LvArray::ChaiBuffer>
```

So it is dropping the redundant const. If you dry to do:
```
double const const a = 1.0;
```
you get:
```
/usr/WS2/settgast/Codes/geosx/GEOSX_wave/src/main/main.cpp:34:14: error: duplicate 'const' declaration specifier [-Werror,-Wduplicate-decl-specifier]
double const const aScalar = 1.0;
```

so looks like it is illegal, but the compiler drops the second const when it is an alias??

Comment (klevzoff) [reply src/ArrayView.hpp:89]: I think in http://eel.is/c++draft/dcl.type.cv#1 this sentence
```
Each cv-qualifier shall appear at most once in a cv-qualifier-seq.
```
refers to language grammar, i.e. what goes into text of a program, so it prohibits literally typing `const const` anywhere. But then
```
Redundant cv-qualifications are ignored.
[Note [2]: For example, these could be introduced by typedefs. — end note]
```

corbett5: > I had put in `ArrayVIew<T>::operator=( ArrayView<T> const&)`, `ArrayVIew<T>::operator=( ViewTypeConst const &)`, but then thought better of it

We rely on the shallow copy semantics of ArrayView in at least once place where we construct the element accessors. Plus it'd be weird to have the copy constructor do something so different from the assignment operator.

PR #268, opened 2022-09-20
add block for FPE on apple M1 chip

rrsettgast:


src/system.cpp
```diff
@@ -528,10 +528,40 @@ int getDefaultFloatingPointExceptions()
...
+        printf("Code detected : %d\n",fe_code);
+
+    abort();
+}
+
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 int enableFloatingPointExceptions( int const exceptions )
 {
 #if defined(__APPLE__) && defined(__MACH__)
+#if !defined(__x86_64__)
```
rrsettgast, line 555: I don't know why 
```c++
#if defined( __arm__)
```
didn't work

corbett5: Yeah that's weird.

corbett5 (approved):

PR #269, opened 2022-09-22
Allow to execute without access to source directory

XL64:
Required to launch a make test from a compute node that cannot see source directory (typicall case on Pangea3)


corbett5 (approved):
Thanks!

PR #270, opened 2022-09-28
Avoiding cuda compiler error in debug

rrsettgast:


src/indexing.hpp
```diff
@@ -186,12 +186,15 @@ LVARRAY_HOST_DEVICE inline constexpr
...
+  int * const p_curDim = &curDim;
+
   bool invalid = false;
-  typeManipulation::forEachArg( [dims, &curDim, &invalid]( auto const index )
-  {
-    invalid = invalid || ( index < 0 ) || ( index >= dims[ curDim ] );
-    ++curDim;
-  }, indices ... );
+  bool * const p_invalid = &invalid;
+  typeManipulation::forEachArg( [dims, p_curDim, p_invalid] LVARRAY_HOST_DEVICE ( INDEX_TYPE const index )
```
corbett5, line 193: I guess you can't capture by reference in host device lambdas?

corbett5 (approved):
Gotta love how this just now shows up two years later...

src/indexing.hpp
rrsettgast: I guess not. It makes sense as it would cause a race condition?

corbett5: No race condition here. Each CUDA thread creates its own copy of the lambda. Just another silly NVCC restriction that only occasionally shows up.

rrsettgast: *could cause a race condition

PR #271, opened 2022-09-30
Add ubuntu 22.04 to travis host-config change

bmhan12:
This PR:

- Makes clang@10.0.0 compiler on quartz target the gcc@8.1 toolchain instead of gcc@4.9.3 by default to get around a missing std feature error when compiling trilinos.

Related to GEOSX/GEOSX#2090
Related to GEOSX/thirdPartyLibs#200


TotoGaz (approved):
Thx @bmhan12 !

PR #277, opened 2023-01-27
Feature/povolny1/polar decomposition

povolny1:
This PR implements the polar decomposition for real square 2x2 and 3x3 matrices using Higham's iterative algorithm. This is needed for some constitutive models and particularly a new hypoelastic update that I'm implementing.


scripts/uberenv/spack_configs/toss_3_x86_64_ib_python/packages.yaml
```diff
@@ -107,3 +107,10 @@ packages:
     externals:
     - spec: pkg-config@0.27.1
       prefix: /usr/bin/
+  
+  ninja:
```
corbett5, line 111: Sorry, I shouldn't have added this. Can you remove it please?

src/fixedSizeSquareMatrixOpsImpl.hpp
```diff
@@ -529,6 +529,56 @@ struct SquareMatrixOps< 2 >
...
+    // Initialize
+    copy< 2, 2 >( R, matrix );
+    FloatingPoint RInverse[2][2] = { {0} },
+                  RInverseTranspose[2][2] = { {0} },
+                  RRTMinusI[2][2] = { {0} };
+
+    // Higham Algorithm
+    FloatingPoint error = 1.0;
+    int iter = 0;
+    while( error > 1.0e-16 && iter < 100 )
```
corbett5: This seems pretty specific for `double`, maybe change it to `10 * LvArray::numeric_limits< FloatingPoint >::epsilon`

src/fixedSizeSquareMatrixOpsImpl.hpp
```diff
@@ -529,6 +529,56 @@ struct SquareMatrixOps< 2 >
...
+      error = 0.0;
+
+      invert( RInverse, R );
+      transpose< 2, 2 >( RInverseTranspose, RInverse );
+      add< 2, 2 >( R, RInverseTranspose );
+      scale< 2, 2 >( R, 0.5 );
+      Rij_eq_AikBjk< 2, 2, 2 >( RRTMinusI, R, R );
+      addIdentity< 2 >( RRTMinusI, -1.0 );
+
+      for( std::ptrdiff_t i = 0 ; i < 2 ; i++ )
```
corbett5: We don't have a L2-Norm for matrices?

src/fixedSizeSquareMatrixOpsImpl.hpp
```diff
@@ -529,6 +529,56 @@ struct SquareMatrixOps< 2 >
...
+      {
+        for( std::ptrdiff_t j = 0 ; j < 2 ; j++ )
+        {
+          error += RRTMinusI[i][j] * RRTMinusI[i][j];
+        }
+      }
+    }
```
corbett5: Do you want to error out (or at least return a signal) if it doesn't converge?

src/fixedSizeSquareMatrixOpsImpl.hpp
```diff
@@ -1162,6 +1212,56 @@ struct SquareMatrixOps< 3 >
...
+  /**
+   * @brief Determine the polar decomposition of the 3x3 matrix @p matrix
+   * @tparam DST_MATRIX The type of @p R.
+   * @tparam MATRIX The type of @p matrix.
+   * @param R The resultant rotation matrix.
+   * @param matrix The matrix to be decomposed.
+   * @details The polar decomposition returns a rotation matrix such that @p R . U = V . @p R = @p matrix.
+   *   This is done using Higham's iterative algorithm.
+   */
+  template< typename DST_MATRIX, typename MATRIX >
```
corbett5, line 1248: I'm probably just not seeing it, but is there a difference between the algorithm for 2x2 and 3x3 matrices? If not you should probably just write a single function. If it works for any size square matrix then it should be made general and moved out of `fixedSizeSquareMatrixOps`

corbett5 (commented):
Can you add some tests?

src/fixedSizeSquareMatrixOpsImpl.hpp
povolny1: The algo does work for any nonsingular square matrix, but requires an inverse calculation which I believe is only explicitly implemented for 2x2 and 3x3 matrices.

src/fixedSizeSquareMatrixOpsImpl.hpp
povolny1: I don't think so

src/fixedSizeSquareMatrixOpsImpl.hpp
corbett5: Yeah my bad, I missed that. I would still suggest that you write one function to work on both 2x2 and 3x3 matrices.

povolny1: ok, will do

povolny1: So it turns out I'm not sure how to implement this. I moved `polarDecomposition` to the unspecialized version of `SquareMatrixOps`, but then the compiler complains about `polarDecomposition` not being implemented in the specialized versions. It seems this problem is typically overcome using inheritance, but that seems like overkill here. Is there an easier way to do this?

corbett5: Yeah, jut make a free function and then have the specialized versions just be a thin wrapper.

povolny1:
Unit test is implemented and passing, I will make an accompanying PR for GEOSX soon

src/fixedSizeSquareMatrixOpsImpl.hpp
povolny1: What should I use to print an error message? Using LVARRAY_WARNING_IF() fails to compile since it calls std::cout.

corbett5: I'd either return the achieved tolerance or the number of iterations or just a boolean to indicate success or failure. Alternatively you can use `LVARRAY_ERROR_IF` which works on device to just kill the program.

PR #290, opened 2023-05-31
c++17 standard

CusiniM:
Related to PR: https://github.com/GEOS-DEV/GEOS/pull/2496


cmake/CMakeBasics.cmake
```diff
@@ -12,7 +12,7 @@ option( ENABLE_TOTALVIEW_OUTPUT "" OFF )
 set( LVARRAY_BUILD_OBJ_LIBS OFF CACHE BOOL "" )
 
 
-# if( NOT BLT_CXX_STD STREQUAL c++14 )
+# if( NOT BLT_CXX_STD STREQUAL c++17 )
```
wrtobin: I suppressed this on Frontier when I was swapping between standards to debug very Frontier / cce/hipcc specific issues. 

I think we should probably reactivate it to ensure that all our users / downstream host-configs appropriately update to C++17 when this is merged.

corbett5: I don't know. I would say until we actually start using c++17 features in LvArray (which may be as soon as this is merged IDK), I would leave it at the minimum requirement. I had a branch for my own use and this is what I wound up doing

```
if( CMAKE_CXX_STANDARD IN_LIST "98; 11" )
    MESSAGE(FATAL_ERROR "LvArray requires at least c++14")
endif()
```

CusiniM: yeah, I have just changed it locally. I ll push it later.

host-configs/LLNL/lassen-cuda-11-base.cmake
```diff
@@ -0,0 +1,15 @@
+include(${CMAKE_CURRENT_LIST_DIR}/lassen-base.cmake)
+
+# Cuda options
+set(ENABLE_CUDA ON CACHE BOOL "")
+set(CUDA_TOOLKIT_ROOT_DIR /usr/tce/packages/cuda/cuda-11.2.0 CACHE STRING "")
+set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER} CACHE STRING "")
+set(CMAKE_CUDA_COMPILER ${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc CACHE STRING "")
+set(CUDA_ARCHITECTURES sm_70 CACHE STRING "")
+set(CMAKE_CUDA_STANDARD 14 CACHE STRING "")
```
rrsettgast:
```suggestion
set(CMAKE_CUDA_STANDARD 17 CACHE STRING "")
```

host-configs/LLNL/lassen-clang-10.cmake
rrsettgast, line 1: keep the cuda version in the filename

host-configs/LLNL/lassen-clang-13.cmake
rrsettgast, line 1: keep cuda version in filename

host-configs/LLNL/lassen-clang-upstream.cmake
rrsettgast, line 1: I think we just ditch this host compiler

host-configs/LLNL/lassen-cuda-11-base.cmake
```diff
@@ -0,0 +1,15 @@
+include(${CMAKE_CURRENT_LIST_DIR}/lassen-base.cmake)
+
+# Cuda options
+set(ENABLE_CUDA ON CACHE BOOL "")
+set(CUDA_TOOLKIT_ROOT_DIR /usr/tce/packages/cuda/cuda-11.2.0 CACHE STRING "")
```
rrsettgast:
```suggestion
set(CUDA_TOOLKIT_ROOT_DIR /usr/tce/packages/cuda/cuda-11.6.1 CACHE STRING "")
```

CusiniM: do we have 11.6 now? I ll update it on the docker image too then.

wrtobin: Current lassen cuda module list:
image

CusiniM: so should we use the latest cuda11 ?

wrtobin: Looking at the release notes they did some work in 11.7 and 11.8 to reduce both host and device memory footprints, so unless we have a compelling reason not to, I would say go for 11.8 yeah.

wrtobin (approved):

PR #294, opened 2023-06-29
Benchmark compilation issues on lassen / pecan with cpp17 + cuda11

wrtobin:


corbett5 (commented):
So you hand inlined `makeRajaView` and that fixed the problem? Not a huge fan of the solution, but it seems like a nasty problem...

Just curious if you tried changing the return type of `makeRajaView` to `auto` and or changed the variable assigned to from `makeRajaView` to `auto`. You never know what might trip up NVCC and get it to actually work haha.

corbett5:
@wrtobin, I just saw your commit message, I love it.

PR #297, opened 2023-07-18
Fixes for Tioga Compiler Warnings

wrtobin:


src/Macros.hpp
```diff
@@ -123,7 +121,11 @@
  */
 
 #if defined(LVARRAY_DEVICE_COMPILE)
-  #if !defined(NDEBUG) || __HIP_DEVICE_COMPILE__ == 1
+//   #if defined(__HIP_DEVICE_COMPILE__)
```
corbett5, line 124: So for now `LVARRAY_ERROR_IF` is just a no-op on HIP devices? Wouldn't it be better to drop the `print` and just directly abort. I feel like that would still aid in debugging.

corbett5 (approved):

src/Macros.hpp
wrtobin: At the moment is still a no-op in release, since the assert compiles out, yeah. Also we can't build in debug regardless since the kernels blow the device memory trivially in debug, even `-O1 -g`, and it just displays as a compiler failure instead of being nice about it like nvcc.

An abort is probably fine, though IIRC there were issues with older HIP versions / devices where abort would cause the device to hang, so we'll need to evaluate on our current platforms with current versions ( almost certain this has been fixed since rocm@5: )

corbett5: Maybe I'm just not parsing the `#if`s correctly, but you're saying that for HIP in both debug and release `LVARRAY_ERROR_IF` contains an `assert`? I know for CUDA it's an `assert` in debug and an abort + print in `release`.

wrtobin: When we're compiling on device `__HIP_DEVICE_COMPILE__` will always be defined (and ==1), so we'll always hit the `assert` version on HIP right now, regardless of `NDEBUG`.

wrtobin: On host we'll hit the version lower down still though.

wrtobin: Having `printf` on device on HIP is supposed to work but causes all sorts of compilation issues in practice. If we had it active many of our kernels should fail to compile.

corbett5: Sounds good to me, the CUDA `printf` was added later if I recall. If there is a good way to abort I would definitely add that, but that doesn't need to be done now. Just curious, does HIP have a decent debugger? Often times for CUDA I'll just put in a bunch of print statements, but it sounds like that wouldn't work.

wrtobin: roc-gdb is functional on device, but without a viable low-optimization build it is pretty difficult to leverage without educated guessing about what will and won't be optimized out, but not *impossible*.

It was at least sufficient enough for me to discover an earlier version of the cray-clang hip compiler would actually force-inline a virtual function in a situation where it wasn't language-compliant to do so.

At another point we were literally using the memory movement back to host for the matrix and vector to debug intermediate values at different stages of kernel execution.

I guess what I'm saying is debug as much as possible using cuda and if there is still an issue on hip, get ready to bang your head for a while.

PR #299, opened 2023-07-25
add getAvailableMemoryOnHost() and getAvailableMemoryOnDevice()

XL64:


src/system.hpp
```diff
@@ -9,8 +9,8 @@
  * @file system.hpp
  * @brief Contains functions that interact with the system or runtime environment.
  */
-
-#pragma once
```
klevzoff, line 13: All of LvArray is using `#pragma once`, did you not have any issues with it before?

src/system.hpp
```diff
@@ -117,6 +117,18 @@ int disableFloatingPointExceptions( int const exceptions = getDefaultFloatingPoi
  */
 void setFPE();
 
+/**
+ * @brief Retieves current available memory on host
+ * @return the available memory in bytes.
+ */
+size_t getAvailableMemoryOnHost();
```
corbett5, line 124: I think I'd prefer `getAvailableMemory(MemorySpace const space)` instead of two separate functions.

PR #304, opened 2023-09-07
c-style error format

### LINKED ISSUE #303: Rework `Macros.hpp` for devices
The `MSG` argument fo the `LVARRAY_ERROR_IF` is useless when used on a GPU (device):

https://github.com/GEOS-DEV/LvArray/blob/86af5bfb5204d94dd7b1f4bcf98a37b8de9a630c/src/Macros.hpp#L148

For example, the expression in https://github.com/GEOS-DEV/LvArray/blob/86af5bfb5204d94dd7b1f4bcf98a37b8de9a630c/src/ArraySlice.hpp#L56-L57

is printed as `Array Bounds Check Failed: index= << index <<  m_dims[0]= << m_dims[0]` in the stacktraces (unevaluated, hence not helpful).

One could use `printf` or `snprintf` together with `MSG` and `...` variadic macros arguments which can be used as `printf(MSG, __VA_ARGS__);` maybe in combination with `__VA_OPT__`.

tbeltzun:
Fix https://github.com/GEOS-DEV/LvArray/issues/303.

The rationale is that only `printf` is supported on `CUDA` kernels, so we switch to `c-style` format.

### DIFF
--- a/src/Macros.hpp
+++ b/src/Macros.hpp
@@ -17,11 +17,8 @@
 #include "system.hpp"
 
 // System includes
-#include <fstream>
-#include <sstream>
 #include <iostream>
 #include <type_traits>
-
 
 #if defined(LVARRAY_USE_CUDA) || defined(LVARRAY_USE_HIP)
 /// Macro defined when using a device.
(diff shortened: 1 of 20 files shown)


src/Macros.hpp
```diff
@@ -126,54 +123,53 @@
...
+  do                                                                                    \
+  {                                                                                     \
+    if( EXP )                                                                           \
+    {                                                                                   \
+      printf( "***** ERROR\n"                                                           \
+              "***** LOCATION: " LOCATION "\n"                                          \
+              "***** Block: [%u, %u, %u]\n"                                             \
+              "***** Thread: [%u, %u, %u]\n"                                            \
+              "***** Controlling expression (should be false): " STRINGIZE( EXP ) "\n", \
+              "***** MSG: " STRINGIZE( MSG )"\n\n"                                      \
```
klevzoff, line 146: If `MSG` is a quoted format string now, do you still need to `STRINGIZE` it?

klevzoff (commented):
I'm on board with using `printf` for kernel debugging (there really shouldn't be any other use case to print from a kernel). However I'm a little concerned about having to switch all of our error logging to `printf` just because of it. Just think about the amount of changes in GEOS this will cause. It's also quite cumbersome to have to specify an additional format string `"%s"` everywhere while already using advanced formatting via `fmt`. 

I think it's time we had separate macros for in-kernel vs out-of-kernel logging. The rule should be that any function/lambda marked as `DEVICE` or `HOST_DEVICE` is only allowed to use the former. Enforced (hopefully) via a CUDA CI build with debugging enabled.

tbeltzun:
@klevzoff, thanks for the comments.

This PR was a shot in the dark regarding debugging CUDA kernels: I think it just doesn't work and requires a cleaner rework from scratch based on https://github.com/GEOS-DEV/LvArray/issues/303.

I think I'll close this as :
1. I don't plan working on this (postdoc time is limited);
2. it doesn't work as is (tons of compilation warnings with CUDA).

PR #311, opened 2024-01-10
Remove m_singleParameterResizeIndex from ArrayView to reduce memory.

dkachuma:
- `m_singleParameterResizeIndex` is replaced by a constant value hard-coded to 0.
- Two interface methods are removed:
  *  `int ArrayView::getSingleParameterResizeIndex() const`
  * `void Array::setSingleParameterResizeIndex( int const index )`
 
  Not sure if these are used elsewhere but I managed to build geos without these.

There is a loss of some functionality here. In `examples/exampleArray.cpp` for instance, you could do
```
LvArray::Array< int > array( 5, 6 );
array.setSingleParameterResizeIndex( 1 );
array.resize( 3 );
```
which would reshape the array to 5x3. This cannot be done anymore.


src/Array.hpp
```diff
@@ -452,38 +452,26 @@ class Array : public ArrayView< T,
...
+   * @note The default dimension is given by singleParameterResizeIndex.
    */
   LVARRAY_HOST_DEVICE
   void resizeDefault( INDEX_TYPE const newdim, T const & defaultValue )
   { resizeDefaultDimension( newdim, defaultValue ); }
 
   /**
    * @brief Sets the size of the Array to zero and destroys all the values.
-   * @details Sets the size of m_singleParameterResizeIndex to 0 but leaves the
+   * @details Sets the size of singleParameterResizeIndex to 0 but leaves the
```
corbett5: Also change these docs, to reflect that you removed `singleParameterResizeIndex`.

src/ArrayView.hpp
```diff
@@ -759,7 +745,7 @@ class ArrayView
 
   /// this data member specifies the dimension that will be resized as a result of a call to the
   /// single dimension resize method.
-  int m_singleParameterResizeIndex = 0;
+  static constexpr int singleParameterResizeIndex = 0;
```
corbett5: I would remove this all together.

corbett5 (changes requested):

src/python/PyArray.cpp
corbett5, line 1: I'm glad you compiled the python interface, because I forgot this would require some changes here. Thanks!

src/MallocBuffer.hpp
```diff
@@ -131,8 +131,15 @@ class MallocBuffer : public bufferManipulation::VoidBuffer
   {
     LVARRAY_ERROR_IF_NE( space, MemorySpace::host );
 
+#if (defined(__GNUC__) && (__GNUC__ == 8))
```
corbett5, line 134: These seem unrelated, correct?

corbett5 (commented):
Looks good to me! If the tests are passing (the python interface tests as well) it is a welcome change!

src/MallocBuffer.hpp
dkachuma: Yes. I encountered an unrelated compile issue with gcc8 on my machine. Apparently there was a bug with gcc8 which makes it issue a potentially false warning if the argument to `std::malloc` is not checked.

corbett5 (approved):

PR #313, opened 2024-01-22
Removing `new` operator on already allocated memory to test Frontier memory corruption issue.

wrtobin:
This is WIP work on a memory corruption issue on Frontier for one of the finest-scale ECP problems.

We may decide to merge this after removing all string arrays in geos, which IIRC is one of the cases that requires the `new` usage currently in lvarray (though might be misremembering).

### DIFF
--- a/host-configs/ORNL/crusher-cce@13.0.1.cmake
+++ b/host-configs/ORNL/crusher-cce@13.0.1.cmake
@@ -1,0 +1,41 @@
+set(CONFIG_NAME "crusher-cce@13.0.1" CACHE PATH "") 
+
+# Set up the tpls
+set(GEOSX_TPL_ROOT_DIR "/gpfs/alpine/geo127/world-shared/cray-sles15-zen2/cce-13.0.1" CACHE PATH "")
+set(GEOSX_TPL_DIR ${GEOSX_TPL_ROOT_DIR} CACHE PATH "")
+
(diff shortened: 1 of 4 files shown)


corbett5:
Just curious, the problem is using placement new? Things like

```c++
new ( dst ) T( *first );
```

https://github.com/GEOS-DEV/LvArray/blob/6057f598b005db6efd132f60c298ad9c827fe3cd/src/arrayManipulation.hpp#L185C5-L185C29

wrtobin:
We're getting a memory failure on ~~lassen~~ frontier in that routine so this branch is/was just to test if bypassing that usage happened to resolve the issue. So far seems like that wasn't the reason for failure, which is only occurring on one of our largest stretch problems that uses at least 1/4 of the machine IIRC.

This PR isn't intended to be merged unless we fully confirm this is both the issue and the only way to resolve it.

wrtobin:
Its really more me not trusting the hip compiler to handle syntax / usage that isn't super common (as I have experienced numerous times during the HIP port).

corbett5:
So this is a problem on Lassen with CUDA as well?

Yeah I would also suspect CUDA/HIP of not correctly implementing placement new, since I've only seen it a handful of times outside of LvArray.

wrtobin:
Sorry no, just frontier/hip, I was typing that while in a meeting where we were discussing other things on lassen.

corbett5:
Phew, okay that makes me feel better. If it doesn't work on Lassen that's on me. But HIP...

PR #314, opened 2024-01-23
Add shallow pretty printers and tests

TotoGaz:
I needed `LvArray` pretty printers that would match the `STL` pretty printers pattern.
This version of the pretty printers do not show any implementation details and focus on the sole data. As such I've simplifying the existing pretty printers a bit.

I've added unit tests, but I did not activate those unit tests through any CI/CD process (I think that this is another subject).

### DIFF
--- a/scripts/gdb-printers-shallow.py
+++ b/scripts/gdb-printers-shallow.py
@@ -111,7 +111,7 @@ class __ArrayPrinter(LvArrayPrinter):
     def children(self) ->  Iterable[tuple[str, gdb.Value]]:
         d0, ds = self.dimensions[0], self.dimensions[1:]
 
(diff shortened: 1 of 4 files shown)


corbett5 (approved):
Any extra debugging features are great! I trust you that they work since I've never written these before.

TotoGaz:
> Any extra debugging features are great! I trust you that they work since I've never written these before.

Thanks you for the review 😉  I hope everything's OK on your side!

One limitation that both my pretty printers and the old ones have is that they are not able to deal with permutations...
It's not an algorithmic issue, but moving back and forth from `python` to `gdb` brings some pure computer science difficulties I could not overcome. I've added a `warning` message... 🤷

PR #316, opened 2024-02-09
add pragmas to supress warnings associated with compiler bugs

rrsettgast:
after https://github.com/GEOS-DEV/LvArray/pull/311 we get compiler warnings on quartz...as a result of compiler bugs. 

This uses pragmas to avoid those warnings on specific lines of code.

### DIFF
--- a/src/MallocBuffer.hpp
+++ b/src/MallocBuffer.hpp
@@ -130,16 +130,14 @@ public:
   void reallocate( std::ptrdiff_t const size, MemorySpace const space, std::ptrdiff_t const newCapacity )
   {
     LVARRAY_ERROR_IF_NE( space, MemorySpace::host );
-
-#if (defined(__GNUC__) && (__GNUC__ == 8))
-    // This is a workaround for a GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87544
-    std::size_t constexpr maxSize = (NumericLimits< std::size_t >::max / sizeof( T )) >> 1;
-    std::size_t newSpaceSize = math::min( static_cast< std::size_t >(newCapacity), maxSize ) * sizeof( T );
-#else
     std::size_t newSpaceSize = newCapacity * sizeof( T );
-#endif
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Walloc-size-larger-than="
     // TODO: If std::is_trivially_copyable_v< T > then we could use std::realloc.
     T * const newPtr = reinterpret_cast< T * >( std::malloc( newSpaceSize ) );
+#pragma GCC diagnostic pop
+
 
     std::ptrdiff_t const overlapAmount = math::min( newCapacity, size );
     arrayManipulation::uninitializedMove( newPtr, overlapAmount, m_data );
--- a/src/arrayManipulation.hpp
+++ b/src/arrayManipulation.hpp
@@ -298,7 +298,11 @@ void resize( T * const LVARRAY_RESTRICT ptr,
     if( newSize - size > 0 )
     {
       std::size_t const sizeDiff = integerConversion< std::size_t >( newSize - size );
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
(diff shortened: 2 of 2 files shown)


castelletto1 (approved):
Looks good to me, thanks

dkachuma (approved):
Builds on pecan with gcc8. Thanks.

src/arrayManipulation.hpp
```diff
@@ -298,7 +298,11 @@ void resize( T * const LVARRAY_RESTRICT ptr,
     if( newSize - size > 0 )
     {
       std::size_t const sizeDiff = integerConversion< std::size_t >( newSize - size );
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
```
CusiniM, line 302: does this mean that we won't get any warning whenever trying to access arrays out of bounds?

castelletto1: With GCC that will be unfortunately the case.

src/MallocBuffer.hpp
```diff
@@ -130,16 +130,14 @@ class MallocBuffer : public bufferManipulation::VoidBuffer
...
-#if (defined(__GNUC__) && (__GNUC__ == 8))
-    // This is a workaround for a GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87544
-    std::size_t constexpr maxSize = (NumericLimits< std::size_t >::max / sizeof( T )) >> 1;
-    std::size_t newSpaceSize = math::min( static_cast< std::size_t >(newCapacity), maxSize ) * sizeof( T );
-#else
     std::size_t newSpaceSize = newCapacity * sizeof( T );
-#endif
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Walloc-size-larger-than="
```
CusiniM, line 136: they claim that the bug was fixed in GCC 9. Are we sure that it's a compiler bug and not an actual issue?

castelletto1: The problem keeps appearing, for example see [here](https://github.com/ocornut/imgui/issues/5343).

castelletto1: A simple reproducer (for strings) found [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651):

```c++
#include <string>

void f(const long long v)
{
   std::string line;
   line += " " + std::to_string(v);
}
```

```
$ g++ -Wall -O3 -std=c++20 -o test_warning.o -c test_warning.cxx
```

On quartz with `GCC 12.1.1` gives:

```
[...]
    inlined from ‘void f(long long int)’ at test_warning.cpp:6:16:
/collab/usr/global/tools/tce4/packages/gcc/gcc-12.1.1/include/c++/12/bits/char_traits.h:431:56: warning: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ accessing 9223372036854775810 or more bytes at offsets [2, 9223372036854775807] and 1 may overlap up to 9223372036854775813 bytes at offset -3 [-Wrestrict]
  431 |         return static_cast<char_type*>(__builtin_memcpy(__s1, __s2, __n));
```

src/arrayManipulation.hpp
klevzoff: Diagnostic disabling is localized to these specific lines though (where we know it's safe), the rest of the code should be ok

src/MallocBuffer.hpp
rrsettgast: @castelletto1 That is a different warning `-Wrestict`

src/arrayManipulation.hpp
rrsettgast: This is non-optimal. But I don't see what "array" that is out of bounds in this case. Perhaps something under the covers, but I don't know what "array" would be hanging around that could be out of bounds in a `memset` call. @klevzoff Do you have any ideas?

corbett5:
Both have been merged, but might as well link the two comments.

https://github.com/GEOS-DEV/LvArray/pull/311#discussion_r1456479499

PR #317, opened 2024-02-10
Bugfix/fix compiler pragma for clang

rrsettgast:


wrtobin:
Per the clang documentation, it understands both `#pragma GCC` and `#pragma clang` as synonymous:

https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

So I'm not sure this is necessary, was there a warning/error motivating this?

rrsettgast:
> Per the clang documentation, it understands both `#pragma GCC` and `#pragma clang` as synonymous:
> 
> https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas
> 
> So I'm not sure this is necessary, was there a warning/error motivating this?

Ah. It is because `clang` didn't recognize the warning flag `Walloc-size-larger-than=` and I think `Wstringop-overflow=` as well.

wrtobin (approved):

PR #319, opened 2024-04-05
New host configs for LLNL machines

victorapm:
Add new host configs:

- lassen - cuda 12
- ruby - gcc-12 and clang-14
- dane - gcc-12

The new lassen host-config requires cmake >= 3.26


victorapm (commented):
Minor fix

CusiniM (approved):

PR #328, opened 2024-08-29
Made some sorted error checks associated with bounds checking.

corbett5:
Related to

GEOS-DEV/GEOS#3312


CusiniM:
I merged this into https://github.com/GEOS-DEV/LvArray/pull/344

PR #329, opened 2024-09-19
feat: add asinh function to math.

CusiniM:
Adding hyperbolic sine because it's needed in https://github.com/GEOS-DEV/GEOS/pull/3280

### DIFF
--- a/src/math.hpp
+++ b/src/math.hpp
@@ -1065,6 +1065,36 @@ __half2 log( __half2 const x )
 
 #endif
 
+/**
+ * @return asinh of @p x.
+ * @param x the argument.
+ * @note This set of overloads is valid for any numeric type. If @p x is integral it is converted to @c double
+ *   and the return type is double.
+ */
+LVARRAY_HOST_DEVICE LVARRAY_FORCE_INLINE
+float asinh( float const x )
+{
+#if defined(LVARRAY_DEVICE_COMPILE)
+  return ::asinhf( x );
+#else
+  return std::asinh( x );
+#endif
+}
+
+/// @copydoc asinh( float )
+template< typename T >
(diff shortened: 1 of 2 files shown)


src/math.hpp
```diff
@@ -1065,6 +1065,36 @@ __half2 log( __half2 const x )
...
+double asinh( T const x )
+{
+#if defined(LVARRAY_DEVICE_COMPILE)
+  return ::asinh( double( x ) );
+#else
+  return std::asinh( x );
+#endif
+}
+
+#endif
```
CusiniM:
```suggestion
```

CusiniM:
> Adding hyperbolic sine because it's needed in [GEOS-DEV/GEOS#3280](https://github.com/GEOS-DEV/GEOS/pull/3280)

I have not added half precision versions of these because I would have to implement them. I guess that they could be added if really necessary.

unitTests/testMath.cpp
```diff
@@ -199,6 +211,27 @@ TYPED_TEST( TestMath, exponential )
   this->exponential();
 }
 
+using TestComplexMathTypes = ::testing::Types<
```
corbett5, line 221: Is this just to avoid using `__half`?

corbett5 (approved):

corbett5:
> > Adding hyperbolic sine because it's needed in [GEOS-DEV/GEOS#3280](https://github.com/GEOS-DEV/GEOS/pull/3280)
> 
> I have not added half precision versions of these because I would have to implement them. I guess that they could be added if really necessary.

Yeah I only added half precision support to testing out on the Chevron finite difference code, I don't think we have to maintain it for everything.

unitTests/testMath.cpp
CusiniM: yes.

PR #333, opened 2024-09-23
Add `log_level` check to log macro

### LINKED ISSUE #331: Control verbosity of `LVARRAY_LOG`
Would it make sense to add a runtime variable to turn off/on the output of `LVARRAY_LOG`?

https://github.com/GEOS-DEV/LvArray/blob/355e8d987fbf15c9ac07442bda28bc52a8e80480/src/Macros.hpp#L105

This would allow turning off messages appearing in GEOS logs like this:
```
Allocated    64.0 B to the HOST  : LvArray::Array<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer> 
Allocated    64.0 B to the HOST  : LvArray::Array<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer> 
Allocated    64.0 B to the HOST  : LvArray::Array<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, 1, camp::int_seq<long, 0l>, int, LvArray::ChaiBuffer> 
```

which come from:

https://github.com/GEOS-DEV/LvArray/blob/355e8d987fbf15c9ac07442bda28bc52a8e80480/src/ChaiBuffer.hpp#L528-L532

victorapm:
Simple solution to #331 that adds a check on LVARRAY_LOG_LEVEL to LVARRAY_LOG macro

Closes #331

### DIFF
--- a/src/Macros.hpp
+++ b/src/Macros.hpp
@@ -17,6 +17,7 @@
 #include "system.hpp"
 
 // System includes
+#include <cstdlib> // std::getenv
 #include <fstream>
 #include <sstream>
 #include <iostream>
@@ -100,9 +101,24 @@
 #define TYPEOFREF( X ) std::remove_reference_t< decltype( X ) >
 
 /**
- * @brief Print the expression.
+ * @brief Conditionally print the provided expression
+ *
+ * This macro checks the value of the LVARRAY_LOG_LEVEL environment variable.
+ * If the variable is set and its value is a positive integer, the macro prints
+ * the provided expression(s) to std::cout followed by a newline.
+ * Otherwise, no output is produced.
+ *
+ * @param ... Variadic macro parameters representing the expression(s) to print.
  */
-#define LVARRAY_LOG( ... ) std::cout << __VA_ARGS__ << std::endl
+#define LVARRAY_LOG( ... ) \
+  do \
+  { \
+    const char* const log_level = std::getenv("LVARRAY_LOG_LEVEL"); \
+    if (log_level && std::atoi(log_level) > 0) \
+    { \
+      std::cout << __VA_ARGS__ << std::endl; \
+    } \
+  } while(0)
 
 /**
  * @brief Print the expression string along with its value.


corbett5:
Does the command line option in GEOS fix this for you? If so I'd rather not introduce environment variables into the mix.

victorapm:
Thanks, Ben!

I don't think any of these fix the issue for me, unless I'm missing something

https://github.com/GEOS-DEV/GEOS/blob/f9a64bb2d1fe34eab99adc25ea1d4be822eca129/src/coreComponents/mainInterface/initialization.cpp#L109-L128

corbett5:
This is what you want, but this type of message should be disabled by default.
https://github.com/GEOS-DEV/GEOS/blob/f9a64bb2d1fe34eab99adc25ea1d4be822eca129/src/coreComponents/mainInterface/initialization.cpp#L124C7-L124C27

victorapm:
I think it's disabled by default. But, this doesn't work here:

https://github.com/GEOS-DEV/LvArray/blob/355e8d987fbf15c9ac07442bda28bc52a8e80480/src/ChaiBuffer.hpp#L528-L532

which is directly using LVARRAY_LOG.

I see the trace migration option is controlling the chai callbacks:
https://github.com/GEOS-DEV/GEOS/blob/f9a64bb2d1fe34eab99adc25ea1d4be822eca129/src/coreComponents/mainInterface/ProblemManager.cpp#L226-L233

Do we need a more general option that turns on/off LVARRAY_LOG as well?

corbett5:
Those `LVARRAY_LOG` in `ChaiBuffer.hpp` you linked above are part of the CHAI callback. They should only be called when chai call backs are enabled. If you are still getting those messages after disabling trace data migration then something broke.

victorapm:
Thanks for the info! Even with data migration disabled, I’m still seeing the messages. I’m not sure I have enough knowledge to resolve this on my own (other than the fix proposed here, which I tested and confirmed it works). Could you take a look when you have a moment?

corbett5:
If this PR stopped the printing then clearly the callbacks aren't actually getting disabled. I don't have time to help with this but it sounds like a bug in GEOS (or maybe but most likely not CHAI) and should be easy to fix. I'd start by adding `GEOS_ERROR("Uh oh: " << opts.traceDataMigration )` to

https://github.com/GEOS-DEV/GEOS/blob/f9a64bb2d1fe34eab99adc25ea1d4be822eca129/src/coreComponents/mainInterface/ProblemManager.cpp#L228

just to make sure that you are indeed disabling call backs.

victorapm:
I see the problem now.

Chai constructs its `ArrayManager` with `m_callbacks_active` set to true, while we only turn it off in GEOS after a couple of allocs, hence the log messages with "Allocated" that I'm seeing

How about doing `arrayManager->disableCallbacks()` right after here?

https://github.com/GEOS-DEV/LvArray/blob/355e8d987fbf15c9ac07442bda28bc52a8e80480/src/ChaiBuffer.hpp#L41

corbett5:
I don't think we want to do this in LvArray, plus if you put it there then anytime `LvArray::internal::getArrayManager` is called call backs will be disabled even if the user wanted to enable them. I would put it somewhere high up in GEOS (perhaps directly in main? https://github.com/GEOS-DEV/GEOS/blob/f9a64bb2d1fe34eab99adc25ea1d4be822eca129/src/main/main.cpp#L35) and then remove it from `ProblemManager` https://github.com/GEOS-DEV/GEOS/blob/f9a64bb2d1fe34eab99adc25ea1d4be822eca129/src/coreComponents/mainInterface/ProblemManager.cpp#L142

victorapm:
Thanks, Ben! I'm replacing this PR with https://github.com/GEOS-DEV/GEOS/pull/3367

PR #335, opened 2024-10-01
refactor placement-new to avoid gcc-9.4 __builtin_memset bug

wrtobin:


src/arrayManipulation.hpp
```diff
@@ -292,29 +292,20 @@ void resize( T * const LVARRAY_RESTRICT ptr,
...
+      std::memset( reinterpret_cast< void * >( ptr + size ), 0, ( sizeDiff ) * sizeof( T ) );
     }
-  }
-  else
-  {
-    // Use std::size_t so that when GCC optimizes this it doesn't produce sign warnings.
-    for( std::size_t i = size; i < std::size_t( newSize ); ++i )
+    else
     {
-      new ( ptr + i ) T( std::forward< ARGS >( args )... );
```
wrtobin, line 317: ptr + i seems to be what was triggering it, switching to instead using a ptr of type T as the loop variable seems to avoid the bug

wrtobin: I was wondering if in some circumstance with the constructor arguments known as compile-time constants, it was doing enough optimizations to do some calculations down here constexpr (effectively), but something used in the calculation was uninitialized. After looking around, nothing used in the calculation isn't initialized in the boolean constructor chain, soooo idunno.

PR #341, opened 2025-01-17
fix: build error on Lassen.

CusiniM:
On Lassen when building on Debug with cuda11 (any compiler) the following error shows up:

```c++
/usr/WS2/cusini1/geos/GEOS/src/coreComponents/LvArray/src/indexing.hpp(194): error: calling a __host__ function("std::basic_ios<char,     ::std::char_traits<char> > ::~basic_ios()") from a __host__ __device__ function("std::basic_ios<char,     ::std::char_traits<char> > ::~basic_ios [subobject]") is not allowed

/usr/WS2/cusini1/geos/GEOS/src/coreComponents/LvArray/src/indexing.hpp(194): error: calling a __host__ function("std::basic_ios<char,     ::std::char_traits<char> > ::basic_ios()") from a __host__ __device__ function("std::basic_ios<char,     ::std::char_traits<char> > ::basic_ios [subobject]") is not allowed

2 errors detected in the compilation of "/usr/WS2/cusini1/geos/GEOS/src/coreComponents/mesh/FaceElementSubRegion.cpp".
``` 

The error seems to originate in `LvArray` indexing so I have tried to figure out what was triggering it. This PR is here to keep track of my attempts. 

@corbett5 @wrtobin @rrsettgast help is welcome.

### DIFF
--- a/src/indexing.hpp
+++ b/src/indexing.hpp
@@ -206,7 +206,7 @@ bool invalidIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES con
 template< typename INDEX_TYPE, typename ... INDICES >
 LVARRAY_HOST_DEVICE inline
 void checkIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES const ... indices )
-{ LVARRAY_ERROR_IF( invalidIndices( dims, indices ... ), "Invalid indices. " << printDimsAndIndices( dims, indices ... ) ); }
+{ LVARRAY_ERROR_IF( false , "Invalid indices. " ); }
 
 /**
  * @brief Calculate the strides given the dimensions and permutation.
--- a/src/indexing.hpp
+++ b/src/indexing.hpp
@@ -159,20 +159,40 @@ std::string getIndexString( INDEX const index, REMAINING_INDICES const ... indic
  * @param indices A variadic pack of indices.
  */
 template< typename INDEX_TYPE, typename ... INDICES >
-std::string printDimsAndIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES const... indices )
+LVARRAY_HOST_DEVICE
+void printDimsAndIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES const... indices )
 {
   constexpr int NDIM = sizeof ... (INDICES);
-  std::ostringstream oss;
-  oss << "dimensions = { " << dims[ 0 ];
+  printf( "dimensions = { %d", dims[ 0 ] );
   for( int i = 1; i < NDIM; ++i )
   {
-    oss << ", " << dims[ i ];
+    printf( ", %d", dims[ i ] );
   }
+  printf( "}\n");
 
-  oss << " } indices = " << getIndexString( indices ... );
-
-  return oss.str();
+  printf( "   indices = { " );
+  (printf(" %d, ", indices ),...);
+  printf( "}\n");
 }
+
+
+/**
+ * @brief Function to check if an index is invalid
+ * @tparam DIMS_TYPE The integral type used for the dimensions of the space
+ * @tparam INDEX_TYPE The Integral types of the index to check against 
+ * @param dims A pointer to the dimensions of the space.
+ * @param indices the index to check against.
+ * @return whether the index is invalid
+ */
+template< typename DIMS_TYPE, typename INDEX_TYPE >
+LVARRAY_HOST_DEVICE inline constexpr
+bool invalidIndex( DIMS_TYPE const * const LVARRAY_RESTRICT dims, 
+                   int const dimsIndex,
+                   INDEX_TYPE const index )
+{
+  return (index < 0) || (index >= dims[dimsIndex]);
+}
+
 
 /**
  * @tparam INDEX_TYPE The integral type used for the dimensions of the space.
@@ -185,14 +205,9 @@ template< typename INDEX_TYPE, typename ... INDICES >
 LVARRAY_HOST_DEVICE inline constexpr
 bool invalidIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES const ... indices )
 {
-  int curDim = 0;
   bool invalid = false;
-  typeManipulation::forEachArg( [dims, &curDim, &invalid]( auto const index )
-  {
-    invalid = invalid || ( index < 0 ) || ( index >= dims[ curDim ] );
-    ++curDim;
-  }, indices ... );
-
+  int curDim = 0;
+  (invalidIndex(dims, curDim++, indices),...);
   return invalid;
 }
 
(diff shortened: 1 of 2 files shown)


src/indexing.hpp
```diff
@@ -206,7 +206,7 @@ bool invalidIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES con
 template< typename INDEX_TYPE, typename ... INDICES >
 LVARRAY_HOST_DEVICE inline
 void checkIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES const ... indices )
-{ LVARRAY_ERROR_IF( invalidIndices( dims, indices ... ), "Invalid indices. " << printDimsAndIndices( dims, indices ... ) ); }
+{ LVARRAY_ERROR_IF( false , "Invalid indices. " ); }
```
CusiniM: This fixes the error. 

I first tried to only remove `printDimsAndIndices( dims, indices ... )` which returns an std::string and I though ti would not be allowed in here since it's a device function but that fix alone does not solve it. So, the function that seems to trigger the error is invalidIndices which, however, does not contain much... 

```c++
template< typename INDEX_TYPE, typename ... INDICES >
LVARRAY_HOST_DEVICE inline constexpr
bool invalidIndices( INDEX_TYPE const * const LVARRAY_RESTRICT dims, INDICES const ... indices )
{
  int curDim = 0;
  bool invalid = false;
  typeManipulation::forEachArg( [dims, &curDim, &invalid]( auto const index )
  {
    invalid = invalid || ( index < 0 ) || ( index >= dims[ curDim ] );
    ++curDim;
  }, indices ... );

  return invalid;
}
```

corbett5:
This is pretty strange, especially since this has been working for a few years at this point. That stream stuff is macro'd out on device and in debug mode it should expand to this implementation which uses `assert`
https://github.com/GEOS-DEV/LvArray/blob/00e3ead67060d3c3d4b8291d8e4abbf680b87eb3/src/Macros.hpp#L129

It seems like somehow `LVARRAY_DEVICE_COMPILE` is not defined correctly, and that the reason that it compiles when you substitute `LVARRAY_ERROR_IF( false , "Invalid indices. " );` is that the compiler simply throws away the body `if ( false )` condition. But it doesn't seem possible for this to happen so I'm not sure what's going on here.

CusiniM:
> This is pretty strange, especially since this has been working for a few years at this point. That stream stuff is macro'd out on device and in debug mode it should expand to this implementation which uses `assert`
> 
> https://github.com/GEOS-DEV/LvArray/blob/00e3ead67060d3c3d4b8291d8e4abbf680b87eb3/src/Macros.hpp#L129
> 
> It seems like somehow `LVARRAY_DEVICE_COMPILE` is not defined correctly, and that the reason that it compiles when you substitute `LVARRAY_ERROR_IF( false , "Invalid indices. " );` is that the compiler simply throws away the body `if ( false )` condition. But it doesn't seem possible for this to happen so I'm not sure what's going on here.

This is even weirder.. I also thought the issue was the definition of `LVARRAY_DEVICE_COMPILE` but even commenting out the entire body of that macro I was getting the same exact error as if something in the body of that `invalidIndices` function was the issue...

corbett5:
> 
> This is even weirder.. I also thought the issue was the definition of `LVARRAY_DEVICE_COMPILE` but even commenting out the entire body of that macro I was getting the same exact error as if something in the body of that `invalidIndices` function was the issue...

Even if you do `#define LVARRAY_ERROR_IF( EXP, MSG ) ((void) 0)` (basically define it to a no-op in all cases) it doesn't compile?

CusiniM:
> #define LVARRAY_ERROR_IF( EXP, MSG ) ((void) 0)

that fixes it. Probably  because `EXP` does not even get evaluated in that case. However, just this

````c++
 #define LVARRAY_ERROR_IF( EXP, MSG ) \
  do \
  { \
    if( EXP ) \
    { } } while ( false )\
 #endif 
 ````
 triggers the error.

**Comment by corbett5** [comment]
> > #define LVARRAY_ERROR_IF( EXP, MSG ) ((void) 0)
> 
> that fixes it. Probably because `EXP` does not even get evaluated in that case. However, just this
> 
> ```c++
>  #define LVARRAY_ERROR_IF( EXP, MSG ) \
>   do \
>   { \
>     if( EXP ) \
>     { } } while ( false )\
>  #endif 
> ```
> 
> triggers the error.

Hmmm, with the same error about streams? My guess as to what's happening is that somehow, something above this in GEOS is marked as `__host__` inside a `__device__` or `__host__ __device__` function. With these weird errors like this that's usually the problem. Not that it makes it any easier to find...

**Comment by CusiniM** [comment]
@rrsettgast and I spent quite a bit of time on this. The issue seemed to be in the lambda. Seems to be fixed now.

**Review by corbett5 — APPROVED** [review]
I still have misgivings that this is hiding a problem in GEOS, but this is an improvement in that it will print out the indices on device so that's pretty nice. I see you're still making changes, but looks good to me.

File: src/indexing.hpp [hunk 2]
```diff
@@ -151,6 +153,46 @@ std::string getIndexString( INDEX const index, REMAINING_INDICES const ... indic
...
+ * @tparam INDEX_TYPE The integral type to be printed.
+ * @param val the value to be printed
+ */
+template< typename INDEX_TYPE >
+LVARRAY_HOST_DEVICE
+void printIndexValue( INDEX_TYPE const & val )
+{
+  static_assert( std::is_integral_v< INDEX_TYPE >, "INDEX_TYPE must be an integral type." );
+  
+  if constexpr( std::is_same_v<INDEX_TYPE, int> )
```
CusiniM: if anyone is aware of a portable way to do this, I d be happy to know.

src/indexing.hpp
```diff
@@ -20,6 +20,8 @@
 // TPL includes
 #include <RAJA/RAJA.hpp>
 
+#include <inttypes.h>
```
CusiniM:
```suggestion
```

src/indexing.hpp
```diff
@@ -151,6 +151,55 @@ std::string getIndexString( INDEX const index, REMAINING_INDICES const ... indic
...
+/**
+ * @tparam INDEX_TYPE The integral type to be printed.
+ * @param val the value to be printed
+ */
+template< typename INDEX_TYPE >
+LVARRAY_HOST_DEVICE
+void printIndexValue( INDEX_TYPE const & val )
+{
+  using DecayedType = std::decay_t<INDEX_TYPE>;
+  static_assert( std::is_integral_v< DecayedType >, "INDEX_TYPE must be an integral type." );
```
CusiniM: either I have found compiler bug or something really weird is happening coz I keep getting this error on one of the CI jobs:
https://github.com/GEOS-DEV/GEOS/actions/runs/12938898965/job/36090106173 
I don't see how that index could be anything other than an `integral_type`.

CusiniM: Removing the assert and forcing a `static_cast` to a` long long` works. However, I don't know what non integral type the compiler thinks that index is since it can cast it to a long long without any issue. 
I think we can leave it as is for now.

PR #358, opened 2026-08-28
fix: CUDA/HIP unit tests

victorapm:
* Fix CUDA/HIP unit tests and portability issues.
* Add HIP-specific workaround for tensor tests.
* Fix synchronization in asynchronous memory-copy tests.
* Improve numerical tolerances for tensor operations.
* Avoid integer overflow in unit tests.
* Fix buffer destruction for non-trivial types.


victorapm:
@rrsettgast @corbett5 could you take a look at this PR?

castelletto1 (approved):

